PR #54, opened 2023-10-13
ci: change to closeAndRelease

nicklasl:
## This PR

- Makes the maven-central release drop automatically when closed.

_To be merged when we know that the closing is successful._

### DIFF
--- a/.github/workflows/release_please.yaml
+++ b/.github/workflows/release_please.yaml
@@ -75,7 +75,7 @@ jobs:
         env:
           OSSRH_USERNAME: ${{ secrets.OSSRH_USERNAME }}
           OSSRH_PASSWORD: ${{ secrets.OSSRH_PASSWORD }}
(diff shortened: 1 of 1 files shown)


.github/workflows/release_please.yaml
```diff
@@ -75,7 +75,7 @@ jobs:
         env:
           OSSRH_USERNAME: ${{ secrets.OSSRH_USERNAME }}
           OSSRH_PASSWORD: ${{ secrets.OSSRH_PASSWORD }}
-        run: ./gradlew publishToSonatype closeSonatypeStagingRepository --no-daemon --stacktrace
+        run: ./gradlew publishToSonatype closeAndReleaseSonatypeStagingRepository --no-daemon --stacktrace
```
toddbaert, line 78: This makes sense based on my understanding of Sonatype. It may be why you were being asked for a staging ID before.

fabriziodemaria (approved):

PR #64, opened 2023-11-04
feat: Add new error for type mismatch

thomaspoignant:
## This PR
The standard error `TypeMismatch` has no exception in the SDK.
This PR exposes the exception.


OpenFeature/src/main/java/dev/openfeature/sdk/exceptions/OpenFeatureError.kt
```diff
@@ -43,4 +43,11 @@ sealed class OpenFeatureError : Exception() {
             return ErrorCode.PROVIDER_NOT_READY
         }
     }
+
+    class TypeMismatchError(override val message: String = "The value has not the expected type") :
```
beeme1mr, line 47:
```suggestion
    class TypeMismatchError(override val message: String = "The value doesn't match the expected type") :
```

beeme1mr (approved):

PR #75, opened 2023-12-07
feat: Make the eventing provider specific instead of being singleton

vahidlazio:
## This PR
- Make the eventing provider specific instead of being singleton

### DIFF
--- a/OpenFeature/src/main/java/dev/openfeature/sdk/FeatureProvider.kt
+++ b/OpenFeature/src/main/java/dev/openfeature/sdk/FeatureProvider.kt
@@ -1,6 +1,9 @@
 package dev.openfeature.sdk
 
-interface FeatureProvider {
+import dev.openfeature.sdk.events.EventObserver
+import dev.openfeature.sdk.events.ProviderStatus
+
+interface FeatureProvider : EventObserver, ProviderStatus {
     val hooks: List<Hook<*>>
     val metadata: ProviderMetadata
 
--- a/OpenFeature/src/main/java/dev/openfeature/sdk/async/Extensions.kt
+++ b/OpenFeature/src/main/java/dev/openfeature/sdk/async/Extensions.kt
@@ -25,6 +25,14 @@ fun OpenFeatureClient.toAsync(): AsyncClient? {
     }
 }
 
+suspend fun OpenFeatureAPI.setProviderAndWait(
+    provider: FeatureProvider,
+    dispatcher: CoroutineDispatcher
+) {
+    setProvider(provider)
+    provider.awaitReady(dispatcher)
+}
+
 internal fun FeatureProvider.observeProviderReady() = observe<OpenFeatureEvents.ProviderReady>()
     .onStart {
         if (isProviderReady()) {
@@ -32,19 +40,11 @@ internal fun FeatureProvider.observeProviderReady() = observe<OpenFeatureEvents.
         }
     }
 
-suspend fun OpenFeatureAPI.awaitProviderReady(
-    dispatcher: CoroutineDispatcher = Dispatchers.IO
-) {
-    val provider = getProvider()
-    requireNotNull(provider)
-    return provider.awaitProviderReady(dispatcher)
+inline fun <reified T : OpenFeatureEvents> OpenFeatureAPI.observeEvents(): Flow<T>? {
+    return getProvider()?.observe<T>()
 }
 
-fun OpenFeatureAPI.observeEvents(): Flow<OpenFeatureEvents>? {
-    return getProvider()?.observe()
-}
-
-suspend fun FeatureProvider.awaitProviderReady(
+suspend fun FeatureProvider.awaitReady(
     dispatcher: CoroutineDispatcher = Dispatchers.IO
 ) = suspendCancellableCoroutine { continuation ->
     val coroutineScope = CoroutineScope(dispatcher)
--- a/OpenFeature/src/main/java/dev/openfeature/sdk/async/Extensions.kt
+++ b/OpenFeature/src/main/java/dev/openfeature/sdk/async/Extensions.kt
@@ -1,33 +1,50 @@
 package dev.openfeature.sdk.async
 
+import dev.openfeature.sdk.FeatureProvider
+import dev.openfeature.sdk.OpenFeatureAPI
 import dev.openfeature.sdk.OpenFeatureClient
-import dev.openfeature.sdk.events.EventHandler
 import dev.openfeature.sdk.events.OpenFeatureEvents
 import dev.openfeature.sdk.events.observe
 import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.cancel
+import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.onStart
 import kotlinx.coroutines.flow.take
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.suspendCancellableCoroutine
 
-fun OpenFeatureClient.toAsync(dispatcher: CoroutineDispatcher = Dispatchers.IO): AsyncClient {
-    return AsyncClientImpl(this, dispatcher)
+fun OpenFeatureClient.toAsync(): AsyncClient? {
+    val provider = OpenFeatureAPI.getProvider()
+    return provider?.let {
+        AsyncClientImpl(
+            this,
+            it
+        )
+    }
 }
 
-internal fun observeProviderReady(
-    dispatcher: CoroutineDispatcher = Dispatchers.IO
-) = EventHandler.eventsObserver(dispatcher)
-    .observe<OpenFeatureEvents.ProviderReady>()
+internal fun FeatureProvider.observeProviderReady() = observe<OpenFeatureEvents.ProviderReady>()
     .onStart {
-        if (EventHandler.providerStatus().isProviderReady()) {
+        if (isProviderReady()) {
             this.emit(OpenFeatureEvents.ProviderReady)
         }
     }
 
-suspend fun awaitProviderReady(
+suspend fun OpenFeatureAPI.awaitProviderReady(
+    dispatcher: CoroutineDispatcher = Dispatchers.IO
+) {
+    val provider = getProvider()
+    requireNotNull(provider)
+    return provider.awaitProviderReady(dispatcher)
+}
+
+fun OpenFeatureAPI.observeEvents(): Flow<OpenFeatureEvents>? {
+    return getProvider()?.observe()
+}
+
+suspend fun FeatureProvider.awaitProviderReady(
     dispatcher: CoroutineDispatcher = Dispatchers.IO
 ) = suspendCancellableCoroutine { continuation ->
     val coroutineScope = CoroutineScope(dispatcher)
@@ -40,8 +57,7 @@ suspend fun awaitProviderReady(
     }
 
     coroutineScope.launch {
-        EventHandler.eventsObserver()
-            .observe<OpenFeatureEvents.ProviderError>()
+        observe<OpenFeatureEvents.ProviderError>()
             .take(1)
             .collect {
                 continuation.resumeWith(Result.failure(it.error))
--- a/OpenFeature/src/main/java/dev/openfeature/sdk/events/EventHandler.kt
+++ b/OpenFeature/src/main/java/dev/openfeature/sdk/events/EventHandler.kt
@@ -2,7 +2,6 @@ package dev.openfeature.sdk.events
 
 import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.CoroutineScope
-import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.cancelChildren
 import kotlinx.coroutines.flow.Flow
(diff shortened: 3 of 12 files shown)


OpenFeature/src/main/java/dev/openfeature/sdk/async/Extensions.kt
```diff
@@ -13,21 +14,36 @@ import kotlinx.coroutines.flow.take
...
+) = observe<OpenFeatureEvents.ProviderReady>()
     .onStart {
-        if (EventHandler.providerStatus().isProviderReady()) {
+        if (isProviderReady()) {
             this.emit(OpenFeatureEvents.ProviderReady)
         }
     }
 
-suspend fun awaitProviderReady(
+suspend fun OpenFeatureAPI.awaitProviderReady(
```
fabriziodemaria, line 38: I am thinking if `awaitProviderReady` could be moved away from OpenFeatureAPI (the global layer), and rather something like: `OpenFeatureAPI.getClient().awaitReady`

fabriziodemaria (commented):
Overall, I think there is an opportunity to take inspiration from the Java SDK setup (which should also adhere very well to the Specs): https://github.com/open-feature/java-sdk/tree/main

OpenFeature/src/main/java/dev/openfeature/sdk/async/Extensions.kt
vahidlazio: this works too. I give it a try.

OpenFeature/src/main/java/dev/openfeature/sdk/FeatureProvider.kt
```diff
@@ -1,6 +1,8 @@
 package dev.openfeature.sdk
 
-interface FeatureProvider {
+import dev.openfeature.sdk.events.ProviderEventObserver
+
+interface FeatureProvider : ProviderEventObserver {
```
fabriziodemaria, line 5: I think we still miss an API for the Application Layer to be able to observe the Events. `FeatureProvider` is only meant to be consumed by the SDK.

As I hinted and mentioned in other comments in this PR, I think [Client](https://github.com/open-feature/kotlin-sdk/blob/c9eca676d0fd091a6146de2246435f9d52ab3cae/OpenFeature/src/main/java/dev/openfeature/sdk/Client.kt#L3) also needs to implement  the `ProviderEventObserver` interface

OpenFeature/src/main/java/dev/openfeature/sdk/events/EventHandler.kt
```diff
@@ -12,21 +11,18 @@ import kotlinx.coroutines.flow.filterIsInstance
...
 
 interface EventsPublisher {
     fun publish(event: OpenFeatureEvents)
 }
 
-inline fun <reified T : OpenFeatureEvents> EventObserver.observe() = observe(T::class)
+inline fun <reified T : OpenFeatureEvents> ProviderEventObserver.observe() = observe(T::class)
 
-class EventHandler(dispatcher: CoroutineDispatcher) : EventObserver, EventsPublisher, ProviderStatus {
+class EventHandler(dispatcher: CoroutineDispatcher) : ProviderEventObserver, EventsPublisher {
```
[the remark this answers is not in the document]
vahidlazio: the event handler right now (before this PR) is singleton for the whole app life cycle, after this PR, the event handler will be one instance inside each provider. making it provider level instead of app level.

about removing `EventPublisher` interface, we need to have a way to only expose the event observer for observing the events coming from the provider without exposing publishing events function. I believe we need this 2 interfaces in this case, wdyt?

OpenFeature/src/main/java/dev/openfeature/sdk/FeatureProvider.kt
vahidlazio: after this change, each provider comes with EventObserver interface, it means we can observe each provider for the events. OpenFeatureAPI and client have access to the current provider and therefore we can observe events in the app layer as well. If we want to observe OpenFeatureAPI directly we can write a simple extension to just get the current provider which has been set and observe the events from that.

vahidlazio: the extension exists [here ](https://github.com/open-feature/kotlin-sdk/blob/69379d21ea10661d4111aaf4fbe0e02add99c9c4/OpenFeature/src/main/java/dev/openfeature/sdk/async/Extensions.kt#L38C2-L38C2) actually.

OpenFeature/src/main/java/dev/openfeature/sdk/async/Extensions.kt
```diff
@@ -13,21 +14,36 @@ import kotlinx.coroutines.flow.take
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.suspendCancellableCoroutine
 
-fun OpenFeatureClient.toAsync(dispatcher: CoroutineDispatcher = Dispatchers.IO): AsyncClient {
-    return AsyncClientImpl(this, dispatcher)
+fun OpenFeatureClient.toAsync(
+    dispatcher: CoroutineDispatcher = Dispatchers.IO
+): AsyncClient {
+    val provider = OpenFeatureAPI.getProvider()
+    requireNotNull(provider)
```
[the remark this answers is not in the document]
fabriziodemaria: I think this comment should be addressed. The SDK should never surface exception to the application layer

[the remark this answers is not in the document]
vahidlazio: yep, agree.

OpenFeature/src/main/java/dev/openfeature/sdk/FeatureProvider.kt
fabriziodemaria: I think we will need another extension to allow the application layer to consume _any_ event from the SDK. The application should not consumer events from the Provider directly. This can be addressed in a future PR though imo

OpenFeature/src/main/java/dev/openfeature/sdk/events/EventHandler.kt
```diff
@@ -12,21 +11,18 @@ import kotlinx.coroutines.flow.filterIsInstance
...
 
-interface ProviderStatus {
-    fun isProviderReady(): Boolean
-}
-
-interface EventObserver {
+interface ProviderEventObserver {
     fun <T : OpenFeatureEvents> observe(kClass: KClass<T>): Flow<T>
+    fun isProviderReady(): Boolean
 }
```
fabriziodemaria, line 17: Nit picking: I'd prefer to keep the same naming for observer/publisher: `EventObserver` and `EventPublisher` seems a bit cleaner. `isProviderReady` is also not really part of an Observer pattern? I wonder if the previous interface (before this PR) was better. But this is also not a blocking comment on my side 👍

vahidlazio: makes sense. I try to keep it as it was.

OpenFeature/src/main/java/dev/openfeature/sdk/FeatureProvider.kt
vahidlazio: i'll add it in the same PR. after unplugged ;) :D

OpenFeature/src/main/java/dev/openfeature/sdk/events/EventHandler.kt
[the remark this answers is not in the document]
fabriziodemaria: I guess what is confusing me is why we are writing an EventPublisher implementation on the SDK layer, since the SDK is not publishing any event. This makes me think that maybe each Provider should implement `ProviderEventObserver` in the way they prefer, and not depend on the `EventHandler` at all.

That being said, Providers today [do depend on EventHandler](https://github.com/spotify/confidence-openfeature-provider-kotlin/blob/421782a2b5c62b15a8b29e38e1ad29bac035b4a8/Provider/src/main/java/com/spotify/confidence/ConfidenceFeatureProvider.kt#L256), so this PR is not a regression in this sense. Let's tackle this comment thread in a future PR if necessary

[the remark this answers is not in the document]
vahidlazio: valid question, I suppose that using the `EventHandler` is optional, they can still do their own eventing, but this is the way that sdk has some utility .. like we did the hard part for you now just use it .. basically some common utility that can be useful for other providers. maybe in the future this can be in another package, like kotlin-sdk-extensions .. then people can use that package if they need some other functionality .. also async stuff to be moved to there.

OpenFeature/src/main/java/dev/openfeature/sdk/async/Extensions.kt
```diff
@@ -1,33 +1,50 @@
...
             this.emit(OpenFeatureEvents.ProviderReady)
         }
     }
 
-suspend fun awaitProviderReady(
+suspend fun OpenFeatureAPI.awaitProviderReady(
+    dispatcher: CoroutineDispatcher = Dispatchers.IO
+) {
+    val provider = getProvider()
+    requireNotNull(provider)
```
fabriziodemaria, line 39: Let's remove this too. I guess this function would just return right away if no provider is set?

nicklasl: Hm... If no provider is set then `getProvider()` returns `null` and I assume `requireNotNull` will throw an exception. This is not great IMO.

Ideally this suspending function would just hang until a provider is set on `OpenFeatureAPI` and become ready. However I don't know how to solve that. Ideas?

nicklasl: Is it completely crazy to setup a `flow` where we continously poll `getProvider()` until one becomes available?

vahidlazio: i thought about polling .. but isn't it too much maybe?

OpenFeature/src/main/java/dev/openfeature/sdk/events/EventHandler.kt
```diff
@@ -2,29 +2,28 @@ package dev.openfeature.sdk.events
...
 
-interface ProviderStatus {
-    fun isProviderReady(): Boolean
+interface EventObserver {
+    fun observe(): Flow<OpenFeatureEvents>
 }
 
-interface EventObserver {
-    fun <T : OpenFeatureEvents> observe(kClass: KClass<T>): Flow<T>
+interface ProviderStatus {
```
nicklasl, line 17: nit: I would expect an interface such as this to have a `getStatus(): Status` function.

OpenFeature/src/main/java/dev/openfeature/sdk/async/Extensions.kt
vahidlazio: we have 5 options : 
1- not have this function at all.
2- never return from this suspending if no provider set (bad because you won't know why you don't get out of the await function)
3- exception with correct message so you know you ask for this function early
4- poll until provider is set
5- return if provider is null (bad because returning implies the provider is ready which is not in this case)

IMO, we need to choose between option 4 or 3. I chose the option 3 but if that's violating the not throwing exception rule. I can try the polling thing.

beeme1mr: I may be missing some of the context here but would it be possible to return he no-op provider?

vahidlazio: @beeme1mr we updated the extension to be like following instead of waiting for provider in a different line and we are adding the `setProviderAndAwaitReady`.

```
suspend fun OpenFeatureAPI.setProviderAndAwaitReady(
    provider: FeatureProvider,
    dispatcher: CoroutineDispatcher
) {
    OpenFeatureAPI.setProvider(provider)
    provider.awaitReady(dispatcher)
}
```

OpenFeature/src/main/java/dev/openfeature/sdk/async/Extensions.kt
```diff
@@ -1,33 +1,50 @@
...
 import dev.openfeature.sdk.events.observe
 import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.cancel
+import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.onStart
 import kotlinx.coroutines.flow.take
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.suspendCancellableCoroutine
```
nicklasl, line 17: Need to put some proper effort in documenting these API's but I think this can go in a separate PR.

nicklasl (approved):

OpenFeature/src/test/java/dev/openfeature/sdk/TestFeatureProvider.kt
```diff
@@ -0,0 +1,79 @@
+package dev.openfeature.sdk
+
+import dev.openfeature.sdk.events.EventHandler
+import dev.openfeature.sdk.events.OpenFeatureEvents
+import kotlinx.coroutines.CoroutineDispatcher
+
+class TestFeatureProvider(
```
beeme1mr, line 7: Can this be the in-memory provider? We started adding those in to other SDKs. It can be used in tests or for basic use cases.

OpenFeature/src/test/java/dev/openfeature/sdk/helpers/AlwaysBrokenProvider.kt
```diff
@@ -6,7 +6,10 @@ import dev.openfeature.sdk.Hook
 import dev.openfeature.sdk.ProviderEvaluation
 import dev.openfeature.sdk.ProviderMetadata
 import dev.openfeature.sdk.Value
+import dev.openfeature.sdk.events.OpenFeatureEvents
 import dev.openfeature.sdk.exceptions.OpenFeatureError.FlagNotFoundError
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.flow
```
beeme1mr, line 12: This is imported twice.

beeme1mr (approved):

OpenFeature/src/test/java/dev/openfeature/sdk/TestFeatureProvider.kt
toddbaert: In most SDKs there's a testing provider (or a few) that have more testing features related to testing the SDK itself (such as throwing on init to test init logic, etc). That seems more like what this is.

The [in-memory provider](https://openfeature.dev/specification/appendix-a#in-memory-provider) is mostly for outsiders to test things. I think that's a separate issue.

toddbaert (approved):
I'm not much of a Kotlin dev, so I could be missing something, but this looks right to me.

I think this will make implementing `named providers` more simple.

PR #79, opened 2023-12-18
chore!: Artifact name change

### LINKED ISSUE #67: Rename artifact android-sdk
The main goal of this repository is to build a client SDK for Android users.
For now, the artifact is called `Kotlin-sdk` that is compatible with Android only, this could be misleading because Kotlin is used in the backend as well and this SDK is not well suited for this usage.

As discussed in the community meeting, it would be great to rename the artifact `android-sdk`.
We don't want to rename the repository since we may have a kotlin server SDK in the future.

nicklasl:
## This PR
- Renames the final artifact that is published to maven central.
- Restructures the repo so that the gradle sub project is named android, like the artifact it produces.

### Related Issues

Fixes #67


beeme1mr (approved):
Looks good, thanks!

toddbaert (approved):

thomaspoignant (approved):
Awesome thank you !!

PR #83, opened 2023-12-19
feat: downgrade the kotlin version so it's compatible with AGP 7.4

vahidlazio:
see [here](https://developer.android.com/build/kotlin-support) for version compatibility.

### DIFF
--- a/android/build.gradle.kts
+++ b/android/build.gradle.kts
@@ -5,7 +5,7 @@ plugins {
     id("maven-publish")
     id("signing")
     id("org.jlleitschuh.gradle.ktlint")
-    kotlin("plugin.serialization") version "1.9.10"
+    kotlin("plugin.serialization") version "1.8.10"
 }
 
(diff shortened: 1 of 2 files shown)


nickybondarenko (approved):

android/build.gradle.kts
```diff
@@ -5,7 +5,7 @@ plugins {
     id("maven-publish")
     id("signing")
     id("org.jlleitschuh.gradle.ktlint")
-    kotlin("plugin.serialization") version "1.9.10"
+    kotlin("plugin.serialization") version "1.8.0"
```
nicklasl, line 8: Previously we had this: https://github.com/open-feature/kotlin-sdk/commit/04717a192d4d54f2af9e289a5aaaca26da83798c

nicklasl: we had this set to `1.8.10`

nicklasl (approved):

PR #84, opened 2023-12-20
feat: add getProviderStatus() function

### LINKED ISSUE #62: Provider status management
As mentioned in the [specification](https://openfeature.dev/specification/sections/providers#requirement-242) :

>  The provider MAY define a status field/accessor which indicates the readiness of the provider, with possible values `NOT_READY`, `READY`, `STALE`, or `ERROR`.
Providers without this field can be assumed to be ready immediately.
The diagram below illustrates the possible states and transitions of the status fields.

Currently, the SDK does not offer the providers does take care of anything to indicate the readiness of the provider.
I was wondering if it was on purpose?

nicklasl:
## This PR
- Adds functionality to access the current status of the provider. 

### Related Issues

Fixes #62

### DIFF
--- a/android/src/main/java/dev/openfeature/sdk/NoOpProvider.kt
+++ b/android/src/main/java/dev/openfeature/sdk/NoOpProvider.kt
@@ -65,5 +65,7 @@ class NoOpProvider(override val hooks: List<Hook<*>> = listOf()) : FeatureProvid
 
     override fun isProviderReady(): Boolean = true
 
+    override fun getProviderStatus(): OpenFeatureEvents = OpenFeatureEvents.ProviderReady
+
     data class NoOpProviderMetadata(override val name: String?) : ProviderMetadata
 }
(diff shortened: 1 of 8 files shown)


android/src/main/java/dev/openfeature/sdk/NoOpProvider.kt
```diff
@@ -65,5 +65,7 @@ class NoOpProvider(override val hooks: List<Hook<*>> = listOf()) : FeatureProvid
 
     override fun isProviderReady(): Boolean = true
```
vahidlazio, line 66: should we make this an extension instead which reads the `getProviderStatus` and returns `getProviderStatus() == OpeFeatureEvents.ProviderReady`?

nicklasl: Sounds good. I'll fix!

vahidlazio (approved):

PR #88, opened 2024-01-08
fix: setProviderAndWait does not hang on ProviderError

fabriziodemaria:
## This PR
`setContextAndWait` shouldn't hang if the Provider fails the initalization phase for some reason.

### Related Issues
N/A

### Follow-up Tasks
Similar logic will need to be added to setEvaluationContext 

### How to test
Added unit tests

### DIFF
--- a/android/src/main/java/dev/openfeature/sdk/OpenFeatureAPI.kt
+++ b/android/src/main/java/dev/openfeature/sdk/OpenFeatureAPI.kt
@@ -14,7 +14,7 @@ object OpenFeatureAPI {
         try {
             provider.initialize(context)
         } catch (e: Throwable) {
-            // TODO What to do here?
+            // This is not allowed to happen
         }
     }
 
--- a/android/src/main/java/dev/openfeature/sdk/OpenFeatureAPI.kt
+++ b/android/src/main/java/dev/openfeature/sdk/OpenFeatureAPI.kt
@@ -11,7 +11,11 @@ object OpenFeatureAPI {
     fun setProvider(provider: FeatureProvider, initialContext: EvaluationContext? = null) {
         this@OpenFeatureAPI.provider = provider
         if (initialContext != null) context = initialContext
-        provider.initialize(context)
+        try {
+            provider.initialize(context)
+        } catch (e: Throwable) {
+            // TODO What to do here?
+        }
     }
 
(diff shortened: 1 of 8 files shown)


android/src/main/java/dev/openfeature/sdk/OpenFeatureAPI.kt
```diff
@@ -11,7 +11,11 @@ object OpenFeatureAPI {
     fun setProvider(provider: FeatureProvider, initialContext: EvaluationContext? = null) {
         this@OpenFeatureAPI.provider = provider
         if (initialContext != null) context = initialContext
-        provider.initialize(context)
+        try {
+            provider.initialize(context)
+        } catch (e: Throwable) {
+            // This should never happen
```
nicklasl, line 17: Can we rephrase this to 
> This is not allowed to happen.

and maybe reference the spec?

fabriziodemaria: I don't think there is anything about this in the Spec. It's this specific implementation that expects the Provider to never throw and rather reads ProviderError events from the Provider.
The Java SDK, for example, works in the opposite way and expects the Provider to throw [in case of errors](https://github.com/open-feature/java-sdk/blob/1b6e16017783aa45ece0c7a6b5ba0f01100813c9/src/main/java/dev/openfeature/sdk/ProviderRepository.java#L140C16-L140C16)

android/src/test/java/dev/openfeature/sdk/helpers/SlowProvider.kt
```diff
@@ -0,0 +1,96 @@
...
+import kotlinx.coroutines.flow.flowOf
+import kotlinx.coroutines.launch
+
+class SlowProvider(override val hooks: List<Hook<*>> = listOf(), private var dispatcher: CoroutineDispatcher) : FeatureProvider {
+    override val metadata: ProviderMetadata = SlowProviderMetadata("Slow provider")
+    private var ready = false
+    private var eventHandler = EventHandler(dispatcher)
+    override fun initialize(initialContext: EvaluationContext?) {
+        CoroutineScope(dispatcher).launch {
+            delay(10000)
```
nicklasl, line 25: nice!

nicklasl (approved):

PR #89, opened 2024-01-09
feat: Update observeEvents() API

fabriziodemaria:
## This PR
Update the `OpenFeatureAPI.observeEvents()` API:
- It can be called by the Application layer before any provider is configured, and it will still receive events once a new provider is configured and emitting events
  - Thanks to this, we can now recommend setting up the observation before `setProvider` is called, thus ensuring no event from the configured provider is lost due to raciness between `observer` and `setProvider`

Usage example (from Application layer):
```
viewModelScope.launch {
    OpenFeatureAPI.observeEvents().collect {
        println(">> Event received: $it")
    }
}

viewModelScope.launch {
    OpenFeatureAPI.setProviderAndWait(
        ConfidenceFeatureProvider.create(
            app.applicationContext,
            clientSecret
        ),
        Dispatchers.IO,
        ctx
    )
}

### DIFF
--- a/android/src/main/java/dev/openfeature/sdk/OpenFeatureAPI.kt
+++ b/android/src/main/java/dev/openfeature/sdk/OpenFeatureAPI.kt
@@ -1,15 +1,21 @@
 package dev.openfeature.sdk
 
+import kotlinx.coroutines.flow.MutableSharedFlow
+import kotlinx.coroutines.flow.SharedFlow
+
 @Suppress("TooManyFunctions")
 object OpenFeatureAPI {
     private var provider: FeatureProvider? = null
     private var context: EvaluationContext? = null
+    private val providersFlow: MutableSharedFlow<FeatureProvider> = MutableSharedFlow(replay = 1)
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: android/src/main/java/dev/openfeature/sdk/OpenFeatureAPI.kt [hunk 1]
```diff
@@ -1,15 +1,21 @@
...
     private var context: EvaluationContext? = null
+    private val providersFlow: MutableSharedFlow<FeatureProvider> = MutableSharedFlow(replay = 1)
+    internal val sharedProvidersFlow: SharedFlow<FeatureProvider> get() = providersFlow
 
     var hooks: List<Hook<*>> = listOf()
         private set
 
     fun setProvider(provider: FeatureProvider, initialContext: EvaluationContext? = null) {
         this@OpenFeatureAPI.provider = provider
+        providersFlow.tryEmit(provider)
```
fabriziodemaria, line 18: Note to self: shall we close the flow on `shutdown`? Shall we emit something else on `clearProvider`?

vahidlazio (approved):

PR #90, opened 2024-01-10
docs: Update Events docs and signature

fabriziodemaria:
## This PR
- Rename API signature for Events
- Updated README, including marking Events as done ✅

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,12 +55,12 @@ dependencies {
 ### Usage
 
 ```kotlin
-// configure a provider and get client
 coroutineScope.launch(Dispatchers.IO) {
+  // configure a provider, wait for it to complete its initialization tasks
   OpenFeatureAPI.setProviderAndWait(customProvider)
   val client = OpenFeatureAPI.getClient()
 
-// get a bool flag value
+  // get a bool flag value
   client.getBooleanValue("boolFlag", default = false)
 }
 ```
@@ -74,7 +74,7 @@ coroutineScope.launch(Dispatchers.IO) {
 | ✅      | [Hooks](#hooks)                 | Add functionality to various stages of the flag evaluation life-cycle.                                                             |
 | ❌      | Logging                         | Integrate with popular logging packages.                                                                                           |
 | ❌      | Named clients                   | Utilize multiple providers in a single application.                                                                                |
-| ⚠️      | [Eventing](#eventing)           | React to state changes in the provider or flag management system.                                                                  |
+| ✅️      | [Eventing](#eventing)           | React to state changes in the provider or flag management system.                                                                  |
 | ✅      | [Shutdown](#shutdown)           | Gracefully clean up a provider during application shutdown.                                                                        |
 | ⚠️      | [Extending](#extending)         | Extend OpenFeature with custom providers and hooks.                                                                                |
 
@@ -89,8 +89,10 @@ If the provider you're looking for hasn't been created yet, see the [develop a p
 Once you've added a provider as a dependency, it can be registered with OpenFeature like this:
 
 ```kotlin
-OpenFeatureAPI.setProvider(MyProvider())
+OpenFeatureAPI.setProviderAndWait(MyProvider())
 ```
+
+_(Asynchronous API that doesn't wait is also available)_
 
 
 ### Targeting
@@ -138,8 +140,8 @@ Some providers support additional events, such as `PROVIDER_CONFIGURATION_CHANGE
 Please refer to the documentation of the provider you're using to see what events are supported.
 
 ```kotlin
-OpenFeatureAPI.observeEvents<OpenFeatureEvents.ProviderReady>()
-    ?.collect {
+OpenFeatureAPI.addHandler<OpenFeatureEvents.ProviderReady>()
+    .collect {
         // do something once the provider is ready
     }
 ```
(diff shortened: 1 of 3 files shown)


android/src/test/java/dev/openfeature/sdk/DeveloperExperienceTests.kt
```diff
@@ -95,7 +96,7 @@ class DeveloperExperienceTests {
         val dispatcher = StandardTestDispatcher(testScheduler)
         var eventCount = 0
         CoroutineScope(dispatcher).launch {
-            OpenFeatureAPI.observeEvents().collect {
+            OpenFeatureAPI.addHandler<OpenFeatureEvents.ProviderReady>().collect {
```
fabriziodemaria, line 99: The intention behind this change is to bring the API design as close as possible to the [Spec](https://openfeature.dev/docs/reference/concepts/events/#event-handlers)

README.md
```diff
@@ -210,6 +212,13 @@ class NewProvider(override val hooks: List<Hook<*>>, override val metadata: Meta
         // add necessary changes on context change
     }
 
+    override fun observe(): Flow<OpenFeatureEvents> {
+        // return the observation flow
+    }
```
fabriziodemaria, line 217: This is a Kotlin specific solution that is not on the Spec (i.e. the SDK listening on Events from the Provider in order to perform some logic). Shall we document here what are the expected Events that every Provider will have to implement?

README.md
```diff
@@ -210,6 +227,13 @@ class NewProvider(override val hooks: List<Hook<*>>, override val metadata: Meta
...
+
+    override fun getProviderStatus(): OpenFeatureEvents {
+        // return the event representative of the current Provider Status
```
fabriziodemaria, line 235: We should introduce actual [Status types](https://openfeature.dev/specification/types/#provider-status), rather than re-using Events.

README.md
```diff
@@ -137,13 +139,28 @@ Some providers support additional events, such as `PROVIDER_CONFIGURATION_CHANGE
...
+viewModelScope.launch {
+  OpenFeatureAPI.addHandler<OpenFeatureEvents.ProviderReady>().collect {
+    println(">> ProviderReady event received")
+  }
+}
+
+viewModelScope.launch {
+  OpenFeatureAPI.setProviderAndWait(
+    ConfidenceFeatureProvider.create(
+      app.applicationContext,
```
nicklasl, line 153: nit
```suggestion
      applicationContext,
```

README.md
```diff
@@ -137,13 +139,28 @@ Some providers support additional events, such as `PROVIDER_CONFIGURATION_CHANGE
...
+      clientSecret
+    ),
+    Dispatchers.IO,
+    myContext
```
nicklasl, line 157: nit
```suggestion
    myEvaluationContext
```

README.md
```diff
@@ -137,13 +139,28 @@ Some providers support additional events, such as `PROVIDER_CONFIGURATION_CHANGE
...
+  )
+}
 ```
 
+_(It's only possible to observe events from the global `OpenFeatureAPI`, until multiple providers are supported)_
```
nicklasl, line 162: Maybe we can leave this as a comment in the readme?

README.md
```diff
@@ -210,6 +227,13 @@ class NewProvider(override val hooks: List<Hook<*>>, override val metadata: Meta
         // add necessary changes on context change
     }
 
+    override fun observe(): Flow<OpenFeatureEvents> {
+        // return the observation flow
```
nicklasl, line 231:
```suggestion
        // return a `Flow` of the Events
```

README.md
nicklasl: For sure! Creating an issue that references this.

android/src/main/java/dev/openfeature/sdk/async/Extensions.kt
```diff
@@ -48,9 +48,9 @@ internal fun FeatureProvider.observeProviderReady() = observe<OpenFeatureEvents.
 Observe events from currently configured Provider.
  */
 @OptIn(ExperimentalCoroutinesApi::class)
-fun OpenFeatureAPI.observeEvents(): Flow<OpenFeatureEvents> {
+internal inline fun <reified T : OpenFeatureEvents> OpenFeatureAPI.addHandler(): Flow<T> {
```
nicklasl, line 51: IMO the name of this function (combined with what it does) does not spark joy.
I'll give it some more thought but my immediate reaction is that we should embrace the platform and the conventions instead. 
Of course this should then be highlighted in the readme/documentation.

fabriziodemaria: I am also conflicted. Would you then prefer to leave `observe()`? I'd be happy with that

nicklasl: Maybe we can have both so we don't need to decide 😸?

fabriziodemaria: Pulling in @vahidlazio for an extra opinion

fabriziodemaria: Agreed! And PR updated

README.md
fabriziodemaria: PR has been opened for a while, let's prioritize getting this one merge and think of more improvements in future PRs

PR #94, opened 2024-01-18
refactor!: remove kotlinx.serialization dependency

### LINKED ISSUE #82: Too opinionated when it comes to serialization?
Currently the Android SDK contains a dependency to `kotlinx-serialization` for json and defines Serializers for the [Value class](https://github.com/open-feature/kotlin-sdk/blob/main/android/src/main/java/dev/openfeature/sdk/Value.kt#L78). 

This is perhaps too opinionated and something that should be controlled by the Providers.

I would suggest an attempt to extract the kotlinx serializers and annotations etc to a separate artifact.

nicklasl:
## This PR

- Removes the built in dependency on `kotlinx.serialization` from the android sdk and instead introduces a separate module that has that dependency.

### Related Issues

Fixes #82 

### Notes

The idea is that the provider creator would pick their own serialization solution and not be forced to get a dependency on the `kotlinx.serialization`.


kotlinxserialization/proguard-rules.pro
```diff
@@ -0,0 +1,21 @@
+# Add project specific ProGuard rules here.
+# You can control the set of applied configuration files using the
+# proguardFiles setting in build.gradle.
+#
+# For more details, see
+#   http://developer.android.com/guide/developing/tools/proguard.html
```
vahidlazio, line 6: do we need this file?

kotlinxserialization/src/main/java/dev/openfeature/sdk/kotlinxserialization/File.kt
```diff
@@ -0,0 +1 @@
+package dev.openfeature.sdk.kotlinxserialization
```
vahidlazio, line 1: i think this can be removed.

vahidlazio (approved):

vahidlazio:
@nicklasl can we change the title though? we are only removing the serialization stuff here.

nicklasl:
> @nicklasl can we change the title though? we are only removing the serialization stuff here.

Good point. I chose `refactor` now. Not sure what is correct and also we want it in the changelog etc.

PR #111, opened 2024-11-17
feat: Flag metadata

### LINKED ISSUE #63: Flag metadata implementation
As mentioned in the [specification](https://openfeature.dev/specification/sections/flag-evaluation#requirement-1414) : 

> If the flag metadata field in the flag resolution structure returned by the configured provider is set, the evaluation details structure's flag metadata field MUST contain that value. Otherwise, it MUST contain an empty record.
>
> This flag metadata field is intended as a mechanism for providers to surface additional information about a feature flag (or its evaluation) beyond what is defined within the OpenFeature spec itself. The primary consumer of this information is a provider-specific hook.

The actual version of the SDK does not allow the providers to return **flag metadata**.
I guess it should be an extra field in the [ProviderEvaluation](https://github.com/open-feature/kotlin-sdk/blob/main/OpenFeature/src/main/java/dev/openfeature/sdk/ProviderEvaluation.kt) and it should be forwarded in the [FlagEvaluationDetails](https://github.com/open-feature/kotlin-sdk/blob/main/OpenFeature/src/main/java/dev/openfeature/sdk/FlagEvaluationDetails.kt).

nicklasl:
## This PR

- adds flag metadata to flag evaluation according to [spec](https://openfeature.dev/specification/sections/flag-evaluation#requirement-1414).

### Related Issues

Fixes #63


nickybondarenko (approved):

android/src/main/java/dev/openfeature/sdk/FlagEvaluationMetadata.kt
```diff
@@ -0,0 +1,63 @@
...
+
+    fun getString(key: String): String? = values[key] as? String
+
+    fun getBoolean(key: String): Boolean? = values[key] as? Boolean
+
+    fun getInt(key: String): Int? = values[key] as? Int
+
+    fun getDouble(key: String): Double? = values[key] as? Double
+
+    fun getAny(key: String): Any? = values[key]
```
thomaspoignant, line 13: Following the [specification](https://openfeature.dev/specification/sections/providers#requirement-2210) I don't think we need this `getAny` in the `EvaluationMetadata`.

thomaspoignant (commented):
It looks good to me, I just have a question about the `getAny()` and why do we need it?

android/src/main/java/dev/openfeature/sdk/FlagEvaluationMetadata.kt
nicklasl: Ah! I think this is just a remnant from an attempt I made with an operator function (but I failed).
I'll push an update that removes it.

PR #112, opened 2024-11-17
refactor: use the noop-provider instead of null provider

nicklasl:
## This PR

- instead of having a nullable provider in the API, use the NoOp Provider for that state.

### Notes

May be some more improvements that can be done due to this?


nicklasl:
Should we consider this a breaking change due to the return type is no longer nullable?

nickybondarenko (approved):

beeme1mr (approved):

PR #129, opened 2025-03-03
feat: add otel compatible telemetry signal

### LINKED ISSUE #125: Add OpenTelemetry-Compatible Telemetry Utility Function
## Description

Add a telemetry utility function to the SDK that can be used within the [finally stage](https://openfeature.dev/specification/sections/hooks/#requirement-442) of a hook to create OpenTelemetry-compatible telemetry signals. It must comply with the OpenTelemetry specification and meet the technical requirements, including those from [Appendix D](https://openfeature.dev/specification/appendix-d/) of the OpenFeature Specification and the [OpenTelemetry Semantic Conventions for Feature Flags](https://opentelemetry.io/docs/specs/semconv/feature-flags/feature-flags-logs/).

## Requirements

- Must comply with the OpenTelemetry (OTel) specification.
- Must be included in the SDK.
- Must not include any third-party dependencies.
- Must adhere to the observability conformance requirements in Appendix D of the OpenFeature Specification.

## Example implementation

A JavaScript implementation can found [here](https://github.com/open-feature/js-sdk/pull/1120).

chrfwow:
## This PR
Adds a function to create an otel compatible telemetry signal

### Related Issues

Fixes #125


aepfli (approved):

android/api/android.api
```diff
@@ -6,13 +6,6 @@ public abstract interface class dev/openfeature/sdk/BaseEvaluation {
 	public abstract fun getVariant ()Ljava/lang/String;
 }
 
-public final class dev/openfeature/sdk/BuildConfig {
-	public static final field BUILD_TYPE Ljava/lang/String;
-	public static final field DEBUG Z
-	public static final field LIBRARY_PACKAGE_NAME Ljava/lang/String;
-	public fun <init> ()V
-}
-
```
nicklasl, line 15: The CI check complains about the removal of this.
Was this done manually or automatically removed by `./gradlew apiDump`?

chrfwow: I did not manually touch this file, all changes to it were done by the `apiDump`. Is there something I can do about it?

nicklasl: Sorry, haven't had time to properly look at this but I wonder if it can have something to do with which Gradle version being used.
Did you run this with the gradle wrapper for this project? (aka `./gradlew`)

chrfwow: I did use the gradle wrapper. However, I had to locally upgrade to gradle version 8.9, update the `com.android.library` plugin to 8.3.2, and comment out `android.disableAutomaticComponentCreation=true` in `gradle.properties` to make the project build. I did not commit those changes.

beeme1mr: Reopening to make sure a follow issue is created.

nicklasl: It's [here](https://github.com/open-feature/kotlin-sdk/issues/131).

beeme1mr:
Hey @nicklasl, the [link issue](https://github.com/open-feature/kotlin-sdk/issues/125) includes a bit more context. Basically, this is a way to easily access an OTel compliant structure that can be used in various ways. That's why we decided to start with a helper function that could live in the SDK itself. We can then create different hooks that folks can use to capture the data however they'd like.

It would live in a sperate package potentially but I'm personally fine included it in the SDK itself. FYI @toddbaert @thomaspoignant @lukas-reining @kinyoklion

android/src/main/java/dev/openfeature/sdk/Telemetry.kt
```diff
@@ -0,0 +1,64 @@
...
+
+const val FLAG_EVALUATION_EVENT_NAME = "feature_flag.evaluation"
+
+/**
+ * Creates an [EvaluationEvent] from a flag evaluation. To be used inside a finally hook to provide an OpenTelemetry
+ * compatible telemetry signal.
+ */
+fun <T> createEvaluationEvent(
+    hookContext: HookContext<T>,
+    providerEvaluation: ProviderEvaluation<T>
```
nicklasl, line 33: From what I can tell, there is no hook that exposes the `ProviderEvaluation`. So this function won't be easily callable from a Hook.
I think what you want to take as input is the `FlagEvaluationDetails<T>`?

I don't know if that's because of the SDK not aligning with the spec or a mistake in this implementation.
Another thing I noticed is that the comment says this should be used in a finally hook, while the current definition of the Hook interface doesn't pass the `details` in the `finallyAfter`.

```kotlin
interface Hook<T> {
  fun before(ctx: HookContext<T>, hints: Map<String, Any>) = Unit
  fun after(ctx: HookContext<T>, details: FlagEvaluationDetails<T>, hints: Map<String, Any>) = Unit
  fun error(ctx: HookContext<T>, error: Exception, hints: Map<String, Any>) = Unit
  fun finallyAfter(ctx: HookContext<T>, hints: Map<String, Any>) = Unit
}
```

chrfwow: I largely followed the [Java implementation](https://github.com/open-feature/java-sdk/pull/1346/files#diff-25d75962dbbcc7cf9f1e6a143be3fda91802026170143eb7cb468c8dac951f74R45) for this function, where the providerEvaluation is used as a parameter.
As for the missing details in `finallyAfter`, those were [added in other sdks](https://github.com/open-feature/java-sdk/pull/1262) some time ago.

nicklasl: Right, thanks for the quick fix! 👍
I also opened an issue for fixing the finallyAfter: https://github.com/open-feature/kotlin-sdk/issues/133

beeme1mr: Yeah, sorry about that. It looks like I missed the Kotlin SDK when creating the finally hook update issues.

android/src/main/java/dev/openfeature/sdk/Telemetry.kt
```diff
@@ -0,0 +1,64 @@
...
+
+    if (flagEvaluationDetails.reason == Reason.ERROR.name) {
+        attributes[TELEMETRY_ERROR_CODE] = flagEvaluationDetails.errorCode ?: ErrorCode.GENERAL
+        flagEvaluationDetails.errorMessage?.let { attributes[TELEMETRY_ERROR_MSG] = it }
+    }
+
+    return EvaluationEvent(
+        FLAG_EVALUATION_EVENT_NAME,
+        attributes,
+        body
```
beeme1mr, line 62: This was renamed to `data` based on feedback from OTel.

https://github.com/open-feature/js-sdk/blob/main/packages/shared/src/telemetry/evaluation-event.ts#L11

aepfli (approved):

PR #134, opened 2025-03-08
fix: ignore cancellation exceptions

nicklasl:
## This PR

- Catches and ignores cancellation exceptions.

### Notes
Since we call cancel manually when new providers or evaluation contexts are set it's a normal thing and it's just annoying if they show up as error statuses.


android/src/main/java/dev/openfeature/sdk/OpenFeatureAPI.kt
```diff
@@ -2,6 +2,7 @@ package dev.openfeature.sdk
 
 import dev.openfeature.sdk.events.OpenFeatureProviderEvents
 import dev.openfeature.sdk.exceptions.OpenFeatureError
+import kotlinx.coroutines.CancellationException
```
nicklasl, line 5: this is just a typealias for the import removed

beeme1mr (approved):
Makes sense, thanks!

PR #148, opened 2025-05-17
Kotlin Multiplatform PoC

### LINKED ISSUE #147: Kotlin Multiplatform Support
Hey Folks,

I think it would be a great addition to this SDK, if it supported [Kotlin Multiplatform](https://www.jetbrains.com/kotlin-multiplatform/), which would allow targeting not only Android, but also iOS, Web and native desktop/server apps.

Is the KMP support on the roadmap? I'm happy to contribute if there's interest.

bencehornak:
I was curious how much effort it would be to extend this SDK to multiple KMP platforms. It was surprisingly straight-forward, there are only a few references to the JDK in the project.

## This PR

- Converts the project from an Android project to a [Kotlin Multiplatform](https://www.jetbrains.com/kotlin-multiplatform/) one with the following targets:
  - android
  - jvm
  - linuxX64 (probably this is probably quite niche, we can remove this to not overcommit to new platforms)
  - js (browser & nodejs)
  - ... (iOS?)
- Replace dependency on `java.util.Date` with `kotlin.time.Instant` and `Value.Date` with `Value.Instant` (breaking change)

### Related Issues

Fixes #147

### Follow-up Tasks

- [ ] Discuss  [which platforms](https://www.jetbrains.com/help/kotlin-multiplatform-dev/supported-platforms.html) should be supported -> I think feature flags are mostly used in client-side apps, I was looking for some stats about what the market share of each KMP platform is, but haven't found any statistics about this yet. I'm personally interested in the Android, iOS and JVM targets.
- [ ] CI infrastructure for each supported platform
- [ ] Some providers should be available for them to make this library useful

### How to test

`./gradlew allTests` should run the tests for each platform. For the js/browser platform I have used Chromium, but you can switch to another browser by changing the `useKarma` block in `shared/build.gradle`.


thomaspoignant:
Hey thanks for this, I will look at it as soon as I have a bit of time.
When it comes to server, the dynamic paradigm is really different and it requires a different sdk.

I have played myself in kotlin with the java server sdk and it works well.

README.md
```diff
@@ -48,14 +48,14 @@ Installation via Maven Central is preferred, using the following dependency:
...
-    api("dev.openfeature:android-sdk:0.4.1")
+    api("dev.openfeature:kotlin-sdk:0.4.1")
 }
 ```
 
 ### Usage
 
 ```kotlin
-coroutineScope.launch(Dispatchers.IO) {
+coroutineScope.launch(Dispatchers.Default) {
```
realdadfish, line 58: IO-bound work should be kept on the IO dispatcher (nice explanation here: https://stackoverflow.com/questions/59039991/difference-between-usage-of-dispatcher-io-and-default)

kotlin-sdk/build.gradle.kts
```diff
@@ -10,24 +10,65 @@ plugins {
...
 
+group = project.extra["groupId"].toString()
+version = releaseVersion
+
+kotlin {
+    androidTarget {
+        compilations.all {
+            compileTaskProvider.configure {
+                compilerOptions {
+                    jvmTarget.set(JvmTarget.JVM_11)
```
realdadfish, line 21: Might want to bump that to JVM_17 (https://developer.android.com/build/jdks), unless the compatibility is really needed.

realdadfish (commented):
LGTM

README.md
nicklasl: I think the `Dispatchers.IO` is an Android only thing so I assume that's why this change was made. However I do believe we should do a bigger rewrite of the readme if we decide to bring this in.

bencehornak: Correct, the `Dispatchers.IO` is JVM-only, while `Dispatchers.Default` is available on 
all platforms supported by `kotlinx.coroutines` (common, jsAndWasmShared, jvm, native), see the [docs](https://kotlinlang.org/api/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-dispatchers/), so that's why I have replaced `IO` with `Default` here and in `OpenFeatureAPI.kt`.

@nicklasl I'm planning to add new sections to the README about the multiplatform support, for sure, so we can at least add a comment, that on JVM targets `Dispatchers.IO` is recommended.

@realdadfish shall I create an `expect val defaultDispatcher: CoroutineDispatcher` field for platform-specific defaults in `OpenFeatureAPI.kt`?

realdadfish: Ah, actually I did not know that this is JVM-only, TIL!

@bencehornak I wouldn't go so far with a platform-specific default and start with `Dispatchers.Default` then.

nicklasl:
> Some providers should be available for them to make this library useful

Trying to wrap my brains around this... This wouldn't just work with any js provider, right? It has to be a provider that is written to work with this SDK, a provider that is also multiplatform.

bencehornak:
> > Some providers should be available for them to make this library useful
> 
> Trying to wrap my brains around this... This wouldn't just work with any js provider, right? It has to be a provider that is written to work with this SDK, a provider that is also multiplatform.

Actually, it is technically possible to create wrappers around the java and js providers thanks to Kotlin's interop capabilities. The wrapper would proxy calls to the underlying implementation, and would convert back the return values and exceptions to KMP values/exceptions. I have seen something similar in action for OpenTelemetry in the Embrace repo I referred to on Slack in the thread we talked about this topic.

https://cloud-native.slack.com/archives/C080WHYDG5V/p1747593332038039?thread_ts=1747593332.038039&cid=C080WHYDG5V

As I mentioned on Slack, the current roadblock for such Java wrapper is a) the collision of package names and b) the static vs dynamic context paradigm issue you mentioned on Slack.

a) we can easily solve in this PR even, if it's not too many breaking changes for a single PR. Regarding b) my thoughts are that it's incorrect to categorize Java as a server language and Kotlin as a client one, given that both can be used for both purposes. I'd much rather have Providers for both languages that support both modes of operation.

But this is maybe a bit off-topic for this PR, I suggest that we switch to the Slack thread to discuss the question of the wrapper.

kotlin-sdk/build.gradle.kts
bencehornak: @realdadfish setting this to JVM 17 would prevent JVM 11 users from adopting this library, wouldn't it? If so, I'd keep the target for this lib low to maximize compatibility and to not break anyone's toolchain setup.

[Temurin will support JDK 11 until 31 Oct 2027](https://endoflife.date/eclipse-temurin), so for now it's far from being dead.

bencehornak:
> KMP would be only targeting static context (aka client) SDKs anyways right? It would be great to have a SDK working for client apps on all these environments.

@lukas-reining I'm with you, the static-context (aka client) paradigm sounds very reasonable to me as the current focus for the Kotlin SDK. If the community around this repo grows, I could imagine an implementation for the dynamic-context paradigm as well, pretty much like for JS.

> If the rewrite is really that straight forward and the KMP version works for all platforms without additional effort, I think this would be great to get in.
> 
> I would only want to make sure that the effort maintaining the SDK does not dramatically increase if the SDK targets all KMP platforms. I guess we need solid test runs on all of the platforms then right?

To JetBrains's credit so far it's been light sailing, I haven't needed to write a single line of platform-specific code 😄 I believe it's not gonna get particularly complicated for maintaining this SDK, given that it contains more interfaces than actual implementations. The test infra will definitely need to be there, I think the current tests will make it for all platform given that I haven't added any platform-specific stuff.

For the Providers, on the other hand, it might be challenging to support multiple platforms based on how they are implemented:

* If they are just wrapping the Android or iOS feature flag SDK of their backend (similar to the [Flagsmith Java Provider](https://github.com/open-feature/java-sdk-contrib/blob/main/providers/flagsmith/pom.xml#L30-L34)), then bunch of platform-specific code and abstractions over that will be needed, requiring more work and maintenance effort for each additional targeted platform.
* If they are depending directly on some REST/gRPC/... API (similar to the [flagd Java Provider](https://github.com/open-feature/java-sdk-contrib/blob/74774b984a00a84d5db615c2bddfcc3b4ab3926a/providers/flagd/pom.xml#L40-L61)), then it's possible to have an actual multiplatform implementation based on pure Kotlin technologies (such as Ktor or Okio) keeping the overhead of each platform to a minimum.

Ultimately, the authors of the Providers can decide, which platform they are targeting, it's not a requirement by any means for them to target all platforms supported by the SDK. The best thing we can do on the SDK side is to support as many platforms as we reasonably can, allowing the Provider authors to target all the platforms they like.

bencehornak:
I'll be AFK for the rest of the week, but I'm planning to come back to this topic next week to make sure that all parts are up to speed before merging the PR. If somebody, who knows, how to create and bootstrap the `kotlin-sdk-contrib` repository in the `open-feature` org could do exactly that, then I could support developing the first pure KMP provider (flagd?) as a PoC with these changes.

realdadfish:
Yes, since AGP 8.0 JDK 17 is a build time requirement.

bencehornak:
Alright, then let me upgrade to JDK17. Is there any other property I need to set to stay compatible with JDK11 projects, despite using a newer SDK?

bencehornak:
> I would also like us to think if we are willing to keep go to some extra length to provide (publishing to maven central) the android-sdk as well as kotlin-sdk?

@nicklasl since this PR is already bringing in some breaking changes (renaming of the base package, the replacement of `java.util.Date` with `kotlin.time.Instant`) the users must unfortunately perform some steps when upgrading anyways. I'd argue that it is better to push through with these breaking changes in a single version than interrupting the library users multiple times (additionally having two up-to-date versions is confusing).

So instead, I'd suggest [_relocating_](https://maven.apache.org/guides/mini/guide-relocation.html) the artifact on Maven Central like this:
image

kotlin-sdk/build.gradle.kts
bencehornak: I would stick to `JVM_11`, see my [comment](https://github.com/open-feature/kotlin-sdk/pull/148#issuecomment-2931141941)

nicklasl:
> > I would also like us to think if we are willing to keep go to some extra length to provide (publishing to maven central) the android-sdk as well as kotlin-sdk?
> 
> @nicklasl since this PR is already bringing in some breaking changes (renaming of the base package, the replacement of `java.util.Date` with `kotlin.time.Instant`) the users must unfortunately perform some steps when upgrading anyways. I'd argue that it is better to push through with these breaking changes in a single version than interrupting the library users multiple times (additionally having two up-to-date versions is confusing).
> 
> So instead, I'd suggest [_relocating_](https://maven.apache.org/guides/mini/guide-relocation.html) the artifact on Maven Central

Yes, I agree. 
I didn't know about the relocating feature.... I asked the AI about it and it claims that as long as you produce the correct pom (you would then produce an extra pom for the old version with the relocation) it "should work".... Then again, that was just an AI... 
<details>
<summary>I tried it out using the code below inside `publishing` and did `publishToMavenLocal` and it seems to be legit.</summary>

```kts
// --- Relocation publication for the OLD 0.4.1 artifact ---
// This will publish a POM for dev.openfeature:android-sdk:0.4.1 that redirects.
create<MavenPublication>("relocationForOldAndroidSdk0_4_1") {
    // Define the OLD coordinates for this publication
    groupId = "dev.openfeature"   // The old groupId
    artifactId = "android-sdk"    // The old artifactId
    version = "0.4.1"             // The specific old version being relocated

    pom.withXml {
        asNode().appendNode("distributionManagement").appendNode("relocation").apply {
            // Point to the NEW coordinates
            appendNode("groupId", "dev.openfeature")
            appendNode("artifactId", "kotlin-sdk")
            // Point to the specific version of the new artifact you want users to use
            // when they try to resolve the old 0.4.1. In your case, this is 0.5.0.
            appendNode("version", "0.5.0")
            appendNode("message", "This artifact has been relocated. Please update your dependency to dev.openfeature:kotlin-sdk:0.5.0.")
        }
    }
}
```
</details>

chrfwow (approved):

nicklasl:
I'm good with shipping this as is. 
One thing that we could improve is to add a part about migration in the readme:
> update dep
the dependency is renamed to dev.openfeature:kotlin-sdk:x.x.x

> fix imports by adding .kotlin
import dev.openfeature.sdk.EvaluationContext -> import dev.openfeature.kotlin.sdk.EvaluationContext

nicklasl:
> @nicklasl sure! should this be inserted in the changelog too? (not sure how to do that with release please)

I think it can be but I'm not a 100% on how. Basically I think it has to do with how you write the commit message when squash-merging the PR.
@toddbaert @beeme1mr if you know of any prior art here, let us know!

toddbaert:
> > @nicklasl sure! should this be inserted in the changelog too? (not sure how to do that with release please)
> 
> I think it can be but I'm not a 100% on how. Basically I think it has to do with how you write the commit message when squash-merging the PR. @toddbaert @beeme1mr if you know of any prior art here, let us know!

It's just a matter of manually editing the release notes in the PR automatically generated by release please.

If you all are fine with a release immediately after merging this, I'm happy to do the edit and then release. If not, then there's not much of a point in doing the edit now, as it will be overwritten the next time release please updates the changelog/release notes (so basically we should do it just before a release). @nicklasl @bencehornak

bencehornak:
@nicklasl I have added the two migration notes to the README. We can add the same notes to the CHANGELOG when preparing release 0.6.0 as @toddbaert has suggested.

PR #151, opened 2025-07-25
fix: Immutability of context

### LINKED ISSUE #150: setEvaluationContext does not support mutable maps well
The functionality to update the evaluation context does not trigger when passing the same (mutable) map of attributes to the `setEvaluationContext` and `setEvaluationContextAndWait` functions.

The issue is [here](https://github.com/open-feature/kotlin-sdk/blob/d8b58b48ad1a89ab31e1f1c2d553e6c7c516f7a8/android/src/main/java/dev/openfeature/sdk/OpenFeatureAPI.kt#L172). There should be a proper map diff. 

This composable component can be used to reproduce in the sample app:
```kotlin
Row(
            modifier = Modifier
                .padding(top = 4.dp)
                .fillMaxWidth(),
            horizontalArrangement = Arrangement.spacedBy(8.dp) // Add spacing between buttons
        ) {
            val myMutableMap by remember { mutableStateOf(mutableMapOf<String, Value>()) }
            Button(modifier = Modifier.weight(1f), onClick = {
                val newKey = myMutableMap.entries.size + 1
                myMutableMap["key$newKey"] = Value.String("value$newKey")
                println("Added key$newKey to mutable map: $myMutableMap. Size is now ${myMutableMap.size}")
                OpenFeatureAPI.setEvaluationContext(ImmutableContext("", myMutableMap))
            }) {
                Text("Set EvaluationContext from mutable map", style = MaterialTheme.typography.bodySmall)
            }
        }
```

fabriziodemaria:
### Problem
The ImmutableContext class was accepting mutable maps in its constructor, allowing callers to modify the original map after instantiation, which would affect the "immutable" context. This violated the immutability guarantee and could lead to unexpected behavior.

### Solution
Implemented comprehensive deep immutability through defensive copying at all levels

### Related Issues
Fixes #150

### Notes
This change is backwards compatible at the API level. However if an integration was performing collection mutation for collections used in a context object, the behavior might now change.

### DIFF
--- a/android/src/main/java/dev/openfeature/sdk/Value.kt
+++ b/android/src/main/java/dev/openfeature/sdk/Value.kt
@@ -21,9 +21,41 @@ sealed interface Value {
 
     data class Date(val date: java.util.Date) : Value
 
-    data class Structure(val structure: Map<kotlin.String, Value>) : Value
+    data class Structure private constructor(val structure: Map<kotlin.String, Value>) : Value {
+        companion object {
+            private fun deepCopyValue(value: Value): Value {
+                return when (value) {
+                    is Structure -> Structure(value.structure.mapValues { (_, v) -> deepCopyValue(v) })
+                    is List -> List(value.list.map { deepCopyValue(it) })
+                    is Date -> Date(value.date.clone() as java.util.Date)
+                    else -> value
+                }
+            }
 
-    data class List(val list: kotlin.collections.List<Value>) : Value
+            operator fun invoke(structure: Map<kotlin.String, Value>): Structure {
+                val deepCopiedStructure = structure.mapValues { (_, value) -> deepCopyValue(value) }
+                return Structure(deepCopiedStructure)
+            }
+        }
+    }
+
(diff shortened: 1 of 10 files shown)


android/src/main/java/dev/openfeature/sdk/Value.kt
```diff
@@ -21,9 +21,41 @@ sealed interface Value {
 
     data class Date(val date: java.util.Date) : Value
 
-    data class Structure(val structure: Map<kotlin.String, Value>) : Value
+    data class Structure private constructor(val structure: Map<kotlin.String, Value>) : Value {
```
nicklasl, line 24: Isn't this (making the constructor private) an API breaking change?
Same on the `List` data class..

nicklasl: Never mind... This is to make the companion `operator fun invoke(structure: Map<kotlin.String, Value>): Structure {` work as the constructor.

nicklasl (approved):

beeme1mr (approved):

PR #196, opened 2025-11-22
feat: Add iOS target

PenguinDan:
## This PR
Adds support for the iOS target to the Kotlin SDK.
- Adds `iosArm64`, `iosX64`, and `iosSimulatorArm64` targets to the build configuration.
- Updates the `README.md` to reflect iOS support.

A major use case for a Kotlin SDK is allowing the implementation to be utilized on Kotlin Multiplatform projects, so updating things here to support that use case

### Related Issues
This change enables the SDK to be used in Kotlin Multiplatform Mobile (KMM) projects targeting iOS devices and simulators (both Apple Silicon and Intel).

### Follow-up Tasks
- [ ] specific iOS integration tests

### DIFF
--- a/kotlin-sdk/build.gradle.kts
+++ b/kotlin-sdk/build.gradle.kts
@@ -40,15 +40,11 @@ kotlin {
         }
     }
     linuxX64 {}
-    listOf(
-        iosX64(),
-        iosArm64(),
-        iosSimulatorArm64()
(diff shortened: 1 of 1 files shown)


PenguinDan:
@bencehornak Yeah sure thing, I should have actually just included that from the beginning

PR #225, opened 2026-04-01
feat!: Adding Long type support

maxnrp:
## Intent

Introduce support for `Long` values throughout the Kotlin SDK, extending the existing numeric handling which previously only supported `Int` and `Double`.

## Motivation

Some feature flag providers and use cases require numeric values beyond the range of `Int`. Supporting `Long` ensures better interoperability and prevents overflow or truncation issues when working with large numeric values.

## Changes

Updated the codebase to support `Long` as a first-class numeric type across the SDK.

## Tests

Updated tests to cover and reflect `Long` as a first-class numeric type across the SDK.

### Test Results
```
BUILD SUCCESSFUL in 1s
63 actionable tasks: 2 executed, 61 up-to-date
```

## Breaking changes

- Public API changes
- The need to implement the getLongEvaluation method.


dd-oleksii (approved):

kotlin-sdk/src/commonTest/kotlin/dev/openfeature/kotlin/sdk/ValueTests.kt
```diff
@@ -18,6 +18,12 @@ class ValueTests {
         assertEquals(3, value.asInteger())
     }
 
+    @Test
+    fun testLongShouldConvertToLong() {
+        val value = Value.Long(3L)
```
typotter: I think using a value that's > `MAX_INT` here as well would read better.

kotlin-sdk/src/commonTest/kotlin/dev/openfeature/kotlin/sdk/ProviderSpecTests.kt
```diff
@@ -20,6 +20,9 @@ class ProviderSpecTests {
         val intResult = provider.getIntegerEvaluation("key", 4, ImmutableContext())
         assertNotNull(intResult.value)
 
+        val longResult = provider.getLongEvaluation("key", 4L, ImmutableContext())
```
typotter: Is it worthwhile to test with a number larger than `MAX_INT` here as well to verify that a proper Long is propagating end to end?

typotter (approved):
pending `ktlint` fix

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/Value.kt
```diff
@@ -9,6 +9,7 @@ sealed interface Value {
     fun asString(): kotlin.String? = if (this is String) string else null
     fun asBoolean(): kotlin.Boolean? = if (this is Boolean) boolean else null
     fun asInteger(): Int? = if (this is Integer) integer else null
+    fun asLong(): kotlin.Long? = if (this is Long) long else null
```
nicklasl, line 12: Nit/Discussion:
I wonder if we should be a bit friendlier here? Allowing conversion between integer and long when we can.
Not sure what makes most sense though? - I mean, what should happen if we _cannot_ perform the conversion?

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/FeatureProvider.kt
```diff
@@ -44,6 +44,27 @@ interface FeatureProvider {
     fun getDoubleEvaluation(key: String, defaultValue: Double, context: EvaluationContext?): ProviderEvaluation<Double>
     fun getObjectEvaluation(key: String, defaultValue: Value, context: EvaluationContext?): ProviderEvaluation<Value>
 
+    /**
+     * Providers should implement this properly, following
+     * the patterns of other evaluation methods.
+     * Note: This default will be removed at 1.0 release.
+     */
+    fun getLongEvaluation(key: String, defaultValue: Long, context: EvaluationContext?): ProviderEvaluation<Long> {
```
nicklasl: Not sure how I feel about the default implementation on this API.. I see how it is tempting to make it non breaking, but we are creating a situation where Long support is advertised at the OpenFeature client level while not actually being supported at the provider level. A user calls getLongValue thinking they get Long semantics, but they silently get Int semantics with a Long wrapper.

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/FeatureProvider.kt
```diff
@@ -44,6 +44,27 @@ interface FeatureProvider {
...
+        require(defaultValue in Int.MIN_VALUE..Int.MAX_VALUE) {
```
nicklasl: I would rather make sure we would throw a `OpenFeatureError` here.

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/FeatureProvider.kt
maxnrp: Removed default implementation, I think it is not worth if we are going to have breaking changes in this release, also agree with the fact that we are not actually fully supporting it and may create a misleading path here.

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/Value.kt
maxnrp: In the light of the reasons we are not having a default implementation for `getLongEvaluation`, would be nice to keep the consistency, and callers may define their behavior later.

nicklasl (approved):

PR #226, opened 2026-04-03
feat: In-memory provider (#208)

mstepien:
## InMemoryProvider, Gherkin based tests
This PR introduces the `InMemoryProvider` for the OpenFeature Kotlin SDK and validates its implementation using the official OpenFeature specification.
- Adds the `InMemoryProvider` for managing and evaluating flags entirely in memory, including `ContextEvaluator` and `Flag` models.
- Links the `open-feature/spec` repository as a Git submodule to utilize the official cross-platform test suite.
- Adds End-to-End (E2E) testing suite (`EvaluationSteps`, `GherkinSpecTest`) using Cucumber to verify the new provider adheres to the OpenFeature Gherkin specification. Evaluation targets evaluation.feature file and ignores evaluation_v2.feature.
- Updates the GitHub Actions `ci.yaml` workflow to ensure Git submodules are checked out during CI test runs.

### Related Issues
Fixes #208 

### Notes
Because this PR introduces the `open-feature/spec` as a Git submodule, developers pulling or checking out this branch locally may need to initialize submodules by running:
```bash
git submodule update --init --recursive

### DIFF
--- a/kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/providers/memory/ContextEvaluator.kt
+++ b/kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/providers/memory/ContextEvaluator.kt
@@ -1,0 +1,18 @@
+package dev.openfeature.kotlin.sdk.providers.memory
+
+import dev.openfeature.kotlin.sdk.EvaluationContext
+
+/**
+ * Context evaluator - use for resolving flag according to evaluation context, for handling targeting.
+ */
+fun interface ContextEvaluator<T> {
+    /**
+     * Evaluates the flag's specific variant based on the provided evaluation context.
+     *
+     * @param flag the feature flag representation
+     * @param evaluationContext the context used for targeting
+     * @return the resolved variant key (a string matching a key in the flag's variants map),
+     *         or null if no match was found.
+     */
+    fun evaluate(flag: Flag<T>, evaluationContext: EvaluationContext?): String?
+}
--- a/kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/providers/memory/Flag.kt
+++ b/kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/providers/memory/Flag.kt
@@ -1,0 +1,43 @@
+package dev.openfeature.kotlin.sdk.providers.memory
+
+import dev.openfeature.kotlin.sdk.EvaluationMetadata
+
+/**
+ * Flag representation for the in-memory provider.
+ */
+data class Flag<T>(
+    val variants: Map<String, T>,
+    val defaultVariant: String? = null,
+    val contextEvaluator: ContextEvaluator<T>? = null,
+    val flagMetadata: EvaluationMetadata? = null,
+    val disabled: Boolean = false
+) {
+    init {
+        if (defaultVariant != null && !variants.containsKey(defaultVariant)) {
+            throw IllegalArgumentException("defaultVariant ($defaultVariant) is not present in variants map")
+        }
+    }
+
+    companion object {
+        fun <T> builder() = Builder<T>()
+    }
+
+    class Builder<T> {
+        private val variants = mutableMapOf<String, T>()
+        private var defaultVariant: String? = null
+        private var contextEvaluator: ContextEvaluator<T>? = null
+        private var flagMetadata: EvaluationMetadata? = null
+        private var disabled: Boolean = false
+
+        fun variant(name: String, value: T) = apply { this.variants[name] = value }
+        fun variants(variants: Map<String, T>) = apply { this.variants.putAll(variants) }
+        fun defaultVariant(defaultVariant: String?) = apply { this.defaultVariant = defaultVariant }
+        fun contextEvaluator(contextEvaluator: ContextEvaluator<T>) = apply { this.contextEvaluator = contextEvaluator }
+        fun flagMetadata(flagMetadata: EvaluationMetadata) = apply { this.flagMetadata = flagMetadata }
+        fun disabled(disabled: Boolean) = apply { this.disabled = disabled }
+
+        fun build(): Flag<T> {
+            return Flag(variants.toMap(), defaultVariant, contextEvaluator, flagMetadata, disabled)
+        }
+    }
+}
--- a/kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/providers/memory/InMemoryProvider.kt
+++ b/kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/providers/memory/InMemoryProvider.kt
@@ -1,0 +1,198 @@
+package dev.openfeature.kotlin.sdk.providers.memory
+
+import dev.openfeature.kotlin.sdk.EvaluationContext
+import dev.openfeature.kotlin.sdk.EvaluationMetadata
+import dev.openfeature.kotlin.sdk.FeatureProvider
+import dev.openfeature.kotlin.sdk.Hook
+import dev.openfeature.kotlin.sdk.OpenFeatureStatus
+import dev.openfeature.kotlin.sdk.ProviderEvaluation
+import dev.openfeature.kotlin.sdk.ProviderMetadata
+import dev.openfeature.kotlin.sdk.Reason
+import dev.openfeature.kotlin.sdk.Value
+import dev.openfeature.kotlin.sdk.events.OpenFeatureProviderEvents
+import dev.openfeature.kotlin.sdk.events.OpenFeatureProviderEvents.EventDetails
+import dev.openfeature.kotlin.sdk.events.OpenFeatureProviderEvents.ProviderConfigurationChanged
+import dev.openfeature.kotlin.sdk.events.OpenFeatureProviderEvents.ProviderReady
+import dev.openfeature.kotlin.sdk.exceptions.ErrorCode
+import dev.openfeature.kotlin.sdk.exceptions.OpenFeatureError.FlagNotFoundError
+import dev.openfeature.kotlin.sdk.exceptions.OpenFeatureError.ProviderNotReadyError
+import dev.openfeature.kotlin.sdk.exceptions.OpenFeatureError.TypeMismatchError
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.MutableSharedFlow
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.asSharedFlow
+import kotlinx.coroutines.flow.update
+import kotlin.concurrent.Volatile
+
+class InMemoryProvider(initialFlags: Map<String, Flag<*>> = emptyMap()) : FeatureProvider {
+
+    override val hooks: List<Hook<*>> = emptyList()
+    override val metadata: ProviderMetadata = InMemoryProviderMetadata("InMemoryProvider")
+
+    private val flagsState = MutableStateFlow<Map<String, Flag<*>>>(initialFlags.toMap())
+
+    @Volatile
+    private var state: OpenFeatureStatus = OpenFeatureStatus.NotReady
+
+    private val eventFlow = MutableSharedFlow<OpenFeatureProviderEvents>(extraBufferCapacity = 64)
+
+    override suspend fun initialize(initialContext: EvaluationContext?) {
+        state = OpenFeatureStatus.Ready
+        eventFlow.tryEmit(ProviderReady())
+    }
+
+    override fun shutdown() {
+        state = OpenFeatureStatus.NotReady
+    }
+
+    override suspend fun onContextSet(
+        oldContext: EvaluationContext?,
+        newContext: EvaluationContext
+    ) {
+        // no-op for in-memory provider as flags aren't context-bound on fetch
+    }
+
+    fun updateFlags(newFlags: Map<String, Flag<*>>) {
+        val flagsChanged = newFlags.keys.toSet()
+        flagsState.update { currentFlags ->
+            currentFlags + newFlags
+        }
+        eventFlow.tryEmit(
+            ProviderConfigurationChanged(
+                EventDetails(flagsChanged = flagsChanged, message = "flags changed")
+            )
+        )
+    }
+
+    override fun observe(): Flow<OpenFeatureProviderEvents> = eventFlow.asSharedFlow()
+
+    override fun getBooleanEvaluation(
+        key: String,
+        defaultValue: Boolean,
+        context: EvaluationContext?
+    ): ProviderEvaluation<Boolean> {
+        return getEvaluation(key, defaultValue, context, Boolean::class)
+    }
+
+    override fun getStringEvaluation(
+        key: String,
+        defaultValue: String,
+        context: EvaluationContext?
+    ): ProviderEvaluation<String> {
+        return getEvaluation(key, defaultValue, context, String::class)
+    }
+
+    override fun getIntegerEvaluation(
+        key: String,
+        defaultValue: Int,
+        context: EvaluationContext?
+    ): ProviderEvaluation<Int> {
+        return getEvaluation(key, defaultValue, context, Int::class)
+    }
+
+    override fun getDoubleEvaluation(
+        key: String,
+        defaultValue: Double,
+        context: EvaluationContext?
+    ): ProviderEvaluation<Double> {
+        return getEvaluation(key, defaultValue, context, Double::class)
+    }
+
+    override fun getObjectEvaluation(
+        key: String,
+        defaultValue: Value,
+        context: EvaluationContext?
+    ): ProviderEvaluation<Value> {
+        return getEvaluation(key, defaultValue, context, Value::class)
+    }
+
+    @Suppress("UNCHECKED_CAST")
+    private fun <T : Any> getEvaluation(
+        key: String,
+        defaultValue: T,
+        context: EvaluationContext?,
+        expectedClass: kotlin.reflect.KClass<T>
+    ): ProviderEvaluation<T> {
+        when (val currentState = state) {
+            is OpenFeatureStatus.NotReady -> throw ProviderNotReadyError("provider not yet initialized")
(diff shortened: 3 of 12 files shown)

### CODE REVIEW

File: kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/providers/memory/ContextEvaluator.kt [hunk 1]
```diff
@@ -0,0 +1,10 @@
+package dev.openfeature.kotlin.sdk.providers.memory
+
+import dev.openfeature.kotlin.sdk.EvaluationContext
+
+/**
+ * Context evaluator - use for resolving flag according to evaluation context, for handling targeting.
+ */
+fun interface ContextEvaluator<T> {
+    fun evaluate(flag: Flag<T>, evaluationContext: EvaluationContext?): T?
+}
```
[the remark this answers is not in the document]
mstepien: We have 
`insert_final_newline = false`
at https://github.com/open-feature/kotlin-sdk/blob/main/.editorconfig#L8 
should we turn it on for all?

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/providers/memory/InMemoryProvider.kt
```diff
@@ -0,0 +1,194 @@
...
+
+class InMemoryProvider(initialFlags: Map<String, Flag<*>> = emptyMap()) : FeatureProvider {
+
+    override val hooks: List<Hook<*>> = emptyList()
+    override val metadata: ProviderMetadata = InMemoryProviderMetadata("InMemoryProvider")
+
+    private val flagsState = MutableStateFlow<Map<String, Flag<*>>>(initialFlags.toMap())
+
+    @Volatile
+    var state: ProviderState = ProviderState.NOT_READY
```
typotter: does this need to be public? This will end up in the api surface, so unless we actually need it public, we should avoid adding unnecessarily to the API. Could probably get by with `internal` here

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/providers/memory/InMemoryProvider.kt
```diff
@@ -0,0 +1,194 @@
...
+        }
+
+        val flag = flagsState.value[key] ?: throw FlagNotFoundError(key)
+
+        if (flag.disabled) {
+            return ProviderEvaluation(
+                value = defaultValue,
+                reason = Reason.DISABLED.toString(),
+                metadata = flag.flagMetadata
+                    ?: dev.openfeature.kotlin.sdk.EvaluationMetadata.EMPTY
```
typotter: nit: import `EvaluationMetadata`

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/providers/memory/InMemoryProvider.kt
```diff
@@ -0,0 +1,194 @@
...
+            // Check if expected class is actually instantiated by default, if we reach here there's
+            // a problem
+            throw TypeMismatchError("flag $key value could not be resolved or cast")
+        }
+
+        return ProviderEvaluation(
+            value = value,
+            variant = flag.defaultVariant,
+            reason = reason.toString(),
+            metadata = flag.flagMetadata ?: dev.openfeature.kotlin.sdk.EvaluationMetadata.EMPTY
```
typotter: nit: import

kotlin-sdk/src/jvmTest/kotlin/dev/openfeature/kotlin/sdk/e2e/EvaluationSteps.kt
```diff
@@ -0,0 +1,308 @@
...
+        client = OpenFeatureAPI.getClient()
+    }
+
+    private var booleanFlagValue: Boolean = false
+    private var stringFlagValue: String = ""
+    private var intFlagValue: Int = 0
+    private var doubleFlagValue: Double = 0.0
+    private var objectFlagValue: Value? = null
+
+    private var booleanFlagDetails: dev.openfeature.kotlin.sdk.FlagEvaluationDetails<Boolean>? = null
```
typotter: nit:`import FlagEvaluationDetails`, here and elsewhere

kotlin-sdk/src/jvmTest/kotlin/dev/openfeature/kotlin/sdk/e2e/EvaluationSteps.kt
```diff
@@ -0,0 +1,308 @@
+/* ktlint-disable max-line-length */
```
typotter: prefer to annotate just the lines where line length has to go over the max

kotlin-sdk/src/commonTest/kotlin/dev/openfeature/kotlin/sdk/providers/memory/InMemoryProviderTest.kt
```diff
@@ -0,0 +1,108 @@
...
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.test.runTest
+import kotlin.test.AfterTest
+import kotlin.test.BeforeTest
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertFailsWith
+
+@OptIn(ExperimentalCoroutinesApi::class)
+class InMemoryProviderTest {
```
typotter, line 23: couple of test cases missing:
- disabled flag
- contextEvaluator returning null (fallback to default)
- updateFlags/updateFlag fires `ProviderConfigurationChanged`

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/providers/memory/InMemoryProvider.kt
```diff
@@ -0,0 +1,194 @@
...
+        if (flag.contextEvaluator != null) {
+            try {
+                value =
+                    (flag.contextEvaluator as ContextEvaluator<T>).evaluate(
+                        flag as Flag<T>,
+                        context
+                    )
+                reason = Reason.TARGETING_MATCH
+            } catch (e: Exception) {
+                value = null
```
typotter: catching without doing anything with the exception reason results in a swallowed exception.
Would it make sense to return an error code/message in the result details?

typotter (changes requested):
Solid implementation with a clean `InMemoryProvider` and a nice Flag/Builder API. Nothing major from my pass outside of what has already been flagged/is in discussion.

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/providers/memory/Flag.kt
```diff
@@ -0,0 +1,41 @@
...
+
+        fun variant(name: String, value: T) = apply { this.variants[name] = value }
+        fun variants(variants: Map<String, T>) = apply { this.variants.putAll(variants) }
+        fun defaultVariant(defaultVariant: String) = apply { this.defaultVariant = defaultVariant }
+        fun contextEvaluator(contextEvaluator: ContextEvaluator<T>) = apply { this.contextEvaluator = contextEvaluator }
+        fun flagMetadata(flagMetadata: EvaluationMetadata) = apply { this.flagMetadata = flagMetadata }
+        fun disabled(disabled: Boolean) = apply { this.disabled = disabled }
+
+        fun build(): Flag<T> {
+            val dv = defaultVariant ?: throw IllegalArgumentException("defaultVariant must be set")
```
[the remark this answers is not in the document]
mstepien: Fair enough. I can see your point now in stacking providers and the purpose of InMemoryProvider.

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/providers/memory/InMemoryProvider.kt
```diff
@@ -0,0 +1,194 @@
...
+class InMemoryProvider(initialFlags: Map<String, Flag<*>> = emptyMap()) : FeatureProvider {
+
+    override val hooks: List<Hook<*>> = emptyList()
+    override val metadata: ProviderMetadata = InMemoryProviderMetadata("InMemoryProvider")
+
+    private val flagsState = MutableStateFlow<Map<String, Flag<*>>>(initialFlags.toMap())
+
+    @Volatile
+    var state: ProviderState = ProviderState.NOT_READY
+        private set
```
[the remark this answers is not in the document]
mstepien: reused OpenFeatureStatus

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/providers/memory/InMemoryProvider.kt
```diff
@@ -0,0 +1,196 @@
...
+        if (value == null) {
+            // Check if expected class is actually instantiated by default, if we reach here there's
+            // a problem
+            throw TypeMismatchError("flag $key value could not be resolved or cast")
+        }
+
+        @Suppress("UNCHECKED_CAST")
+        return ProviderEvaluation(
+            value = value as T,
+            variant = if (reason == Reason.DEFAULT || reason == Reason.ERROR) flag.defaultVariant else null,
```
typotter: When there's no `contextEvaluator` and `reason == Reason.STATIC`, this sets `variant = null` even though the resolved value is the `flag.defaultVariant`. The condition looks like it should be inverted — `TARGETING_MATCH` and SPLIT are the cases where the variant is genuinely unknown. Suggest:

```kotlin
variant = if (reason == Reason.TARGETING_MATCH) null else flag.defaultVariant,
```

typotter (commented):
A few things I noticed on the latest commits — great progress overall.

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/providers/memory/InMemoryProvider.kt
```diff
@@ -0,0 +1,196 @@
...
+        }
+
+        if (value != null && !expectedClass.isInstance(value)) {
+            throw TypeMismatchError("flag $key evaluated to a type that does not match expected type")
+        }
+
+        if (value == null) {
+            // Check if expected class is actually instantiated by default, if we reach here there's
+            // a problem
+            throw TypeMismatchError("flag $key value could not be resolved or cast")
```
typotter, line 183: When the evaluator throws AND `defaultVariant` is null (now allowed), `value` stays null after the fallback and the code throws `TypeMismatchError` - discarding the `errorCode` and `errorMessage` that were just set. The caller loses the actual failure reason. In this case would it be more correct to return a `ProviderEvaluation` with the error fields set and `defaultValue` as the value, rather than throwing?

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/providers/memory/InMemoryProvider.kt
```diff
@@ -0,0 +1,196 @@
...
+
+class InMemoryProvider(initialFlags: Map<String, Flag<*>> = emptyMap()) : FeatureProvider {
+
+    override val hooks: List<Hook<*>> = emptyList()
+    override val metadata: ProviderMetadata = InMemoryProviderMetadata("InMemoryProvider")
+
+    private val flagsState = MutableStateFlow<Map<String, Flag<*>>>(initialFlags.toMap())
+
+    @Volatile
+    var state: OpenFeatureStatus = OpenFeatureStatus.NotReady
```
typotter: Still public with `private set`

typotter (commented):
A few paths still don't have unit test coverage:
- The `disabled` flag path (return `defaultValue` with `Reason.DISABLED`)
- `contextEvaluator` returning `null` (fallback to `defaultVariant` with `Reason.DEFAULT`)
- `updateFlags`/`updateFlag` event emission (that `ProviderConfigurationChanged` fires with the correct `flagsChanged` set)

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/providers/memory/ContextEvaluator.kt
```diff
@@ -0,0 +1,10 @@
+package dev.openfeature.kotlin.sdk.providers.memory
+
+import dev.openfeature.kotlin.sdk.EvaluationContext
+
+/**
+ * Context evaluator - use for resolving flag according to evaluation context, for handling targeting.
+ */
+fun interface ContextEvaluator<T> {
+    fun evaluate(flag: Flag<T>, evaluationContext: EvaluationContext?): T?
```
typotter: Should this return the `variantKey` instead of the actual evaluation value? The other SDKs (go, java, js) use context evaluators that return the variant key string that is then used to resolve the variant value from the variants map.

kotlin-sdk/src/commonTest/kotlin/dev/openfeature/kotlin/sdk/providers/memory/InMemoryProviderTest.kt
```diff
@@ -0,0 +1,343 @@
...
+        val targetingEval = localProvider.getBooleanEvaluation("targeting-flag", false, null)
+        assertEquals(Reason.TARGETING_MATCH.toString(), targetingEval.reason)
+        assertEquals(null, targetingEval.variant)
+
+        val defaultEval = localProvider.getBooleanEvaluation("default-flag", false, null)
+        assertEquals(Reason.DEFAULT.toString(), defaultEval.reason)
+        assertEquals("on", defaultEval.variant)
+
+        val errorEval = localProvider.getBooleanEvaluation("error-flag", false, null)
+        assertEquals(Reason.ERROR.toString(), errorEval.reason)
```
typotter, line 247: duplicate assertion on `errorEval.reason`

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/providers/memory/ContextEvaluator.kt
typotter: Under this setup (returning the variant direclty)
- variant is always null for TARGETING_MATCH results — callers can't trace which variant was selected                                                                                                                                                     
- The variants map is bypassed entirely during targeting, so there's no guarantee the returned value corresponds to any defined variant
                                                                                                                                                                                                                                                              
Since ContextEvaluator is a public API, changing the return type later would be a breaking change. Worth aligning with the other SDKs before merging.

typotter (changes requested):
Thanks for the changes, Marcin

Sorry for missing this earlier, but I think we need to relook at the `ContextEvaluator` interface.

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/providers/memory/ContextEvaluator.kt
mstepien: Thanks for bringing that up. ContextEvaluator returns variantKey after this remark.

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/providers/memory/InMemoryProvider.kt
```diff
@@ -0,0 +1,194 @@
...
+            currentFlags + newFlags
+        }
+        eventFlow.tryEmit(
+            ProviderConfigurationChanged(
+                EventDetails(flagsChanged = flagsChanged, message = "flags changed")
+            )
+        )
+    }
+
+    fun updateFlag(flagKey: String, newFlag: Flag<*>) {
```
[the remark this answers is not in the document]
mstepien: Well, specification does not require it. It is present at java-sdk not necessary in other sdks. Let's not keep it until needed.

typotter:
### Code review

Found 2 issues:

1. `kotlin.concurrent.Volatile` is JVM-only — will fail to compile on `linuxX64`, `js`, `iosArm64`, `iosSimulatorArm64`. Replace with `AtomicRef` from `kotlinx-atomicfu` (move from `commonTest` to `commonMain`), or restructure to avoid the field entirely.

https://github.com/open-feature/kotlin-sdk/blob/eaa5335595ee28d14557123356ad1d176df14842/kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/providers/memory/InMemoryProvider.kt#L25-L35

2. `context-aware` E2E fixture: `contextEvaluator` returns `"INTERNAL"`/`"EXTERNAL"` (uppercase) but the variant keys are `"internal"`/`"external"` (lowercase). `containsKey` always misses — every evaluation falls back to `defaultVariant` with `Reason.ERROR` instead of producing `Reason.TARGETING_MATCH`. Fix: return lowercase strings from the evaluator.

https://github.com/open-feature/kotlin-sdk/blob/eaa5335595ee28d14557123356ad1d176df14842/kotlin-sdk/src/jvmTest/kotlin/dev/openfeature/kotlin/sdk/e2e/EvaluationSteps.kt#L41-L47

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

mstepien:
> kotlin-sdk/kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/providers/memory/InMemoryProvider.kt

Volatile should work without changes. And should take effects on platform with multithreading. Project compiles fine on all non-JVM targets without failing. 
Starting with Kotlin 1.9.0, `kotlin.concurrent.Volatile` was introduced into the common standard library and made stable for Kotlin multiplatform.

> kotlin-sdk/kotlin-sdk/src/jvmTest/kotlin/dev/openfeature/kotlin/sdk/e2e/EvaluationSteps.kt

Good catch, fixed case sensitivity issue

PR #228, opened 2026-04-06
feat!: Add hook data (binary breaking change)

maxnrp:
## Intent
Ship hook data on HookContext so hooks can keep per-hook, per-evaluation mutable state across before, after, error, and finallyAfter.

## Motivation
To preserve data throughout the whole lifecycle of the hook stages, allowing to share with with other (observability) tools.

## Changes
- Added **HookData** and internal **DefaultHookData**. Per-hook instances are created on each evaluation; HookContext defaults hookData to an empty store to keep tests and hand-built contexts simple.

- Extended HookContext with hookData: HookData (default empty store for manual construction).

- OpenFeatureClient: builds hooksWithData — one DefaultHookData() per supported hook in merged order for each evaluation.

- HookSupport: invokes hooks with [SupportedType]Ctx.copy(hookData = …) so each hook sees the same HookData across stages and isolated from other hooks.

- Filtering moved from **HookSupport** to **OpenFeatureClient** so we create hooksWithData once: every hook that will run gets its own HookData in lockstep. HookSupport then only dispatches stages with copy(hookData = …), without re-filtering or risking misaligned hook vs. storage lists (the latter only became a possibility after introducing hook data).

- Regenerated JVM/Android .api dumps for the binary compatibility validator.

## Tests
- HookSpecTests: hook data shared across stages for one hook; isolation between two hooks; persistence through error and finally on failed evaluation.
- HookSupportTests: updated for hooksWithData wiring.

### Test Results
```
BUILD SUCCESSFUL in 16s
63 actionable tasks: 63 executed
```

## Breaking changes
Public API changes


dd-oleksii (approved):

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/HookContext.kt
```diff
@@ -6,5 +6,6 @@ data class HookContext<T>(
     val defaultValue: T,
     val ctx: EvaluationContext?,
     val clientMetadata: ClientMetadata?,
-    val providerMetadata: ProviderMetadata
+    val providerMetadata: ProviderMetadata,
+    val hookData: MutableMap<String, Any> = mutableMapOf()
 )
```
typotter, line 11: mind adding a newline while you're here?

kotlin-sdk/src/commonTest/kotlin/dev/openfeature/kotlin/sdk/HookSpecTests.kt
```diff
@@ -76,4 +77,104 @@ class HookSpecTests {
...
+                finallyValue = ctx.hookData["trace"]
+            }
+        }
+
+        client.getBooleanValue("key", false, FlagEvaluationOptions(listOf(hook)))
+
+        assertEquals("from-before", errorValue)
+        assertEquals("from-before", finallyValue)
+    }
 }
```
typotter, line 180: and another here?

typotter (commented):
CI is failing because of a timing issue in the test. 
LGTM aside from the failing test, with a small formatting request

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/HookContext.kt
maxnrp: If I do ktlint fails.

`insert_final_newline = false`
http://github.com/open-feature/kotlin-sdk/blob/main/.editorconfig#L8

kotlin-sdk/src/commonTest/kotlin/dev/openfeature/kotlin/sdk/HookSpecTests.kt
maxnrp: answered above.

typotter (approved):
LGTM. Hook data implementation is clean and well-tested. The StatusTests fix is correct.

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/HookSupport.kt
nicklasl: I was considering if this is a breaking change and I guess it is? (Which means we should mark the PR as such).
But then I questioned why `HookSupport` is public in the first place, maybe it should be internal to the SDK? -- That would also be a breaking change but if we already plan to make it breaking why not do it correctly?

```suggestion
internal class HookSupport {
```

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/HookContext.kt
```diff
@@ -6,5 +6,6 @@ data class HookContext<T>(
     val defaultValue: T,
     val ctx: EvaluationContext?,
     val clientMetadata: ClientMetadata?,
-    val providerMetadata: ProviderMetadata
+    val providerMetadata: ProviderMetadata,
+    val hookData: MutableMap<String, Any> = mutableMapOf()
```
nicklasl: Did we deliberately decide against allowing `null` in this map?
I think it's fine but it differs from how Java handles it IIRC.

nicklasl (changes requested):
My requested changes here is on renaming the PR to clarify it's breaking API (allbeit minor):
suggesting `feat!: Add hook data` (perhaps we can be even more explaining?).

Anyway, if we are marking the PR as breaking we might as well make `HookSupport` internal.

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/HookContext.kt
maxnrp: I believe we can move on allowing null to mirror Java SDK here.

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/HookSupport.kt
```diff
@@ -1,45 +1,44 @@
 package dev.openfeature.kotlin.sdk
 
 @Suppress("UNCHECKED_CAST") // TODO can we do better here?
-class HookSupport {
+internal class HookSupport {
     fun <T> beforeHooks(
         flagValueType: FlagValueType,
         hookCtx: HookContext<T>,
-        hooks: List<Hook<*>>,
+        hooksWithData: List<Pair<Hook<*>, MutableMap<String, Any>>>,
```
toddbaert: I think there was some discussion on using `MutableMap<String, Any?>` to support nulls? I'm not a Kotlin expert so forgive me if I'm misunderstanding.

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/HookSupport.kt
```diff
@@ -1,45 +1,44 @@
...
-            .filter { hook -> hook.supportsFlagValueType(flagValueType) }
-            .forEach { hook ->
+        hooksWithData
+            .asReversed()
+            .forEach { (hook, hookData) ->
                 when (flagValueType) {
                     FlagValueType.BOOLEAN -> {
                         safeLet(hook as? Hook<Boolean>, hookCtx as? HookContext<Boolean>) { booleanHook, booleanCtx ->
-                            booleanHook.before(booleanCtx, hints)
+                            booleanHook.before(booleanCtx.copy(hookData = hookData), hints)
```
toddbaert: Since HookContext is a data class, copy(hookData = hookData) allocates a new object per hook per stage. Could we instead pre-build one HookContext per hook (with its hookData baked in) upfront, and pass the same reference to all stages? That's what both the Java SDK (Pair<Hook, HookContext>) and JS SDK (a hookContexts array, one per hook) do; it avoids the repeated allocations and simplifies HookSupport.

[Java](https://github.com/open-feature/java-sdk/blob/d5f7a645688f5f45129626b5eaac1af32280245e/src/main/java/dev/openfeature/sdk/HookSupport.java)

[JS](https://github.com/open-feature/js-sdk/blob/431f899191a2f7693fea9fab40f7cfb1f22f61f8/packages/server/src/client/internal/open-feature-client.ts#L283-L292)

toddbaert (approved):
Approved on my end, but I left some thoughts as a non-Kotlin dev ❤️ .

And @nicklasl 's approval is definitely required.

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/HookSupport.kt
maxnrp: I've updated it so the context carry the data upfront so we won't the copying in each stage.

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/HookSupport.kt
maxnrp: I believe the idea is that because Java's SDK is not enforcing the HookData to not be null, we should also keep the parity here. And I did that, but, do we actually want that? Kotlin allows null safety with `Any` instead of `Any?`. Was this something that was done on purpose in Java?

...not an expert in any of both languages also.

nicklasl (approved):

PR #230, opened 2026-04-10
feat!: Add StateManagingProvider resource and flow

maxnrp:
### Intent
Address the race condition in the current SDK implementation, where lifecycle state is split between the provider and the API, causing undefined event ordering in concurrent scenarios.

This is achieved by narrowing event emission to a single source of truth — the provider itself — via the `status: StateFlow<OpenFeatureStatus>` property on `StateManagingProvider`, eliminating the window between `initialize()` returning and the API updating its internal state. State transitions are the authoritative origin of events, which are emitted in response to each lifecycle stage change.

### Motivation
[open-feature/spec#365 — Provider lifecycle race condition in multi-threaded SDKs](https://github.com/open-feature/spec/issues/365)
The issue above describes how split ownership of state and events between the API and the provider can produce undefined ordering under concurrency (e.g. API publishing “ready” while the provider emits something else from background work when it actually finishes initializing).

This change aligns the Kotlin SDK with the direction there: the provider is the party that can own a single, authoritative `OpenFeatureStatus` stream. **The API forwards it instead of racing to update the same conceptual state after initialize for implementations**. That does not rewrite the whole spec, but it reduces split ownership for the new path while preserving the old behavior for legacy providers.

Providers that only implement `FeatureProvider` (and not `StateManagingProvider`) keep the existing Provider/SDK-managed path: internal status buffer, mirroring `observe()`, Ready / Error handling around init and context changes, etc.

### Changes

- Introduced `StateManagingProvider` as an extension of `FeatureProvider`.
- `StateManagingProvider` exposes `statusFlow: StateFlow<OpenFeatureStatus>` as the single source of truth for lifecycle state.
- `OpenFeatureAPI` now:
Delegates `getStatus() `and `statusFlow` directly to the provider when it implements `StateManagingProvider`.
Updates `setProvider` and `setProviderAndWait` to suspend until the provider emits a state other than `NotReady` after initialize() completes.
- `NoOpProvider` and `MultiProvider` implement `StateManagingProvider`.

### Tests
**StateManagingProviderStatusTests** — State-managing init, `statusFlow`, `observe` (`ProviderReady`), context changes, swap from legacy to SMP, and `setProviderAndWait` until status leaves `NotReady`.

**LegacyFeatureProviderStatusTests** — Legacy init, observe mirrored into SDK `status`, `onContextSet`, `getStatus` from SDK buffer, provider replacement/shutdown.

**StatusTests** — End-to-end transitions, clear/shutdown, concurrent context updates with a state-managing slow provider, provider replacement (including async `setProvider`).

**MultiProviderTests** — Existing multiprovider behavior (strategies, lifecycle, track, error aggregation, init with long-running flows).

Together, these separate legacy vs state-managing behavior and cover init, status, and context under concurrency where relevant.

### Breaking changes vs legacy compliance
- Intended backward compatible
- `StateManagingProvider` (opt-in for new/custom providers)
- API update: new public type: `StateManagingProvider`


typotter (commented):
.

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/OpenFeatureAPI.kt
```diff
@@ -133,7 +178,11 @@ object OpenFeatureAPI {
         getProvider().shutdown()
         provider = NOOP_PROVIDER
         providersFlow.value = NOOP_PROVIDER
-        _statusFlow.emit(OpenFeatureStatus.NotReady)
+
```
typotter, line 190: There's a bug here: `provider` is already reassigned to `NOOP_PROVIDER` before this check runs. Since `NoOpProvider` now implements `StateManagingProvider`, the condition is always false, so `_statusFlow` never receives `NotReady` when a legacy provider is cleared. 

Fix: capture the old provider before reassignment:
```kotlin
val oldProvider = getProvider()
provider = NOOP_PROVIDER
providersFlow.value = NOOP_PROVIDER
if (oldProvider !is StateManagingProvider) {
    _statusFlow.emit(OpenFeatureStatus.NotReady)
}

File: kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/multiprovider/MultiProvider.kt [hunk 2]
```diff
@@ -95,12 +97,14 @@ class MultiProvider(
     }
 
     private val OpenFeatureStatus.precedence: Int
+        // not sure if I understand this precedence logic
```
typotter: > // not sure if I understand this precedence logic 
this shouldn't ship. ConfigurationChanged at precedence 2 (same as Stale/Reconciling) is reasonable: it's a non-terminal state that doesn't block evaluations the way NotReady/Fatal/Error do. If you're confident in that, remove the comment. If not, consider investigating what precedence other SDKs are using or starting a thread in the slack channel

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/NoOpProvider.kt
```diff
@@ -17,6 +33,8 @@ open class NoOpProvider(override val hooks: List<Hook<*>> = listOf()) : FeatureP
         // no-op
     }
 
+    override fun observe(): Flow<OpenFeatureProviderEvents> = events
```
typotter, line 44: StateManagingProvider$DefaultImpls provides a default observe() that derives events from status. But NoOpProvider overrides observe() with its own MutableSharedFlow, bypassing the default. This means:

- Some implementations derive events from status (via default)
- Others maintain a separate event flow

Callers can't tell which path they're on, and the two can diverge (e.g. status transitions that observe() doesn't mirror).

Recommendation: pick one model and enforce it. Either remove the default observe() and require all SMPs to implement it, or remove the separate events flow from NoOpProvider and rely on the default.

README.md
```diff
@@ -280,10 +280,17 @@ in an Android app.
 ### Develop a provider
 
 To develop a provider, you need to create a new project and include the OpenFeature SDK as a dependency.
-You’ll then need to write the provider by implementing the `FeatureProvider` interface exported by the OpenFeature SDK.
+You’ll then need to write the provider by implementing the `StateManagingProvider` interface exported by the OpenFeature SDK.
```
typotter: The "Develop a provider" section now shows StateManagingProvider as the only starting point, but the PR describes it as opt-in. A developer implementing FeatureProvider directly will think they're doing something wrong.

Suggest adding a note: "FeatureProvider is still fully supported. StateManagingProvider is preferred for new providers that want authoritative status ownership."

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/multiprovider/MultiProvider.kt
```diff
@@ -265,6 +274,7 @@ class MultiProvider(
             }
             throw aggregate
         }
+        _statusFlow.value = OpenFeatureStatus.NotReady
```
typotter, line 275: `_statusFlow.value = OpenFeatureStatus.NotReady` is set here on shutdown, but eventFlow doesn't get a corresponding ProviderNotReady emission. NoOpProvider.shutdown() does call events.tryEmit(ProviderNotReady) — MultiProvider should do the same so observe() callers see the shutdown transition:

typotter (changes requested):
Overall, great work on a rather hard issue:
  - Dual-path design preserves backward compatibility cleanly
  - ConfigurationChanged status addition is needed and sensible
  - ProviderReconciling event fills a real gap
  - Test separation (legacy vs state-managing) is good structure
  - statusFlow alias in MultiProvider preserving binary compat is thoughtful

README.md
maxnrp: I've updated it. Should we also mention the possible buggy behavior with the race condition possibility in the non-SMP option?

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/NoOpProvider.kt
maxnrp: I’ve made `observe()` abstract in SMP, no `emptyFlow()` for SMPs as in legacy, authors are required to implement it.

Just something to think about:
What about automating event emission in the future, as soon as the `status` changes? I think this would make `status`’ changes more sound and less worry for authors (mapping `status` to events on the side). But by doing so, as it is now, would take eventDetails away from authors (there's also a fix for that).

I just think this would be narrowing down a little more the author's burden in keeping a coherent single source of truth, and more complex providers could still override.

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/OpenFeatureAPI.kt
```diff
@@ -130,10 +175,15 @@ object OpenFeatureAPI {
      * Clear the current [FeatureProvider] for the SDK and set it to a no-op provider.
      */
     suspend fun clearProvider() {
-        getProvider().shutdown()
+        val previous = getProvider()
+        previous.shutdown()
         provider = NOOP_PROVIDER
```
typotter: This read and write here should actually hold the mutex on `provider` similar to how we swap in the `setProviderInternal` method.

I know it's not part of your change, but while you're here, mind fixing this?

README.md
typotter: This is a good point to raise. I think that it is worth publicly documenting the race condition so that devs across all SDK communities understand why they will need to make changes in the future.

kotlin-sdk/api/android/kotlin-sdk.api
```diff
@@ -242,6 +242,7 @@ public class dev/openfeature/kotlin/sdk/NoOpProvider : dev/openfeature/kotlin/sd
 	public fun getIntegerEvaluation (Ljava/lang/String;ILdev/openfeature/kotlin/sdk/EvaluationContext;)Ldev/openfeature/kotlin/sdk/ProviderEvaluation;
 	public fun getMetadata ()Ldev/openfeature/kotlin/sdk/ProviderMetadata;
 	public fun getObjectEvaluation (Ljava/lang/String;Ldev/openfeature/kotlin/sdk/Value;Ldev/openfeature/kotlin/sdk/EvaluationContext;)Ldev/openfeature/kotlin/sdk/ProviderEvaluation;
+	public fun getStatus ()Lkotlinx/coroutines/flow/StateFlow;
```
typotter, line 245: Is adding the method here to meet the OpenFeature spec? Or is this required to properly expose the new managed state?

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/StateManagingProvider.kt
```diff
@@ -0,0 +1,25 @@
...
+/**
+ * Manages its own OpenFeatureStatus.
+ *
+ * The SDK reads status directly from `status`. The implementation must update it during
+ * lifecycle events such as initialize, onContextSet, and shutdown.
+ *
+ * Implementations that don’t follow this use the deprecated shared Provider/SDK-managed
+ * status system.
+ */
+interface StateManagingProvider : FeatureProvider {
```
typotter, line 15: **Missing shutdown contract documentation**

Please add docs to `StateManagingProvider.shutdown()` (or the interface-level docs) that implementations need to reset `_status` to `NotReady` here and emit the corresponding event _before_ returning in order to properly sequence the shutdown lifecycle for all clients of `observe`.

kotlin-sdk/src/commonTest/kotlin/dev/openfeature/kotlin/sdk/StateManagingProviderStatusTests.kt
```diff
@@ -0,0 +1,161 @@
...
+
+        OpenFeatureAPI.setProviderAndWait(NoOpProvider())
+        waitAssert { assertTrue(events.isNotEmpty()) }
+
+        job.cancelAndJoin()
+        assertTrue(events.isNotEmpty())
+    }
+
+    @Test
+    fun smp_context_set_updates_status_via_provider() = runTest {
```
typotter, line 81: Consider adding a test for `initialize` throwing or timing out and never emitting `READY`.

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/OpenFeatureAPI.kt
```diff
@@ -39,22 +40,39 @@ object OpenFeatureAPI {
                 tryEmit(OpenFeatureStatus.NotReady)
             }
 
-    val statusFlow: Flow<OpenFeatureStatus> get() = _statusFlow.distinctUntilChanged()
+    /**
+     * Get the current [OpenFeatureStatus] for the SDK.
+     */
+    @OptIn(ExperimentalCoroutinesApi::class)
+    val statusFlow: Flow<OpenFeatureStatus>
+        get() = providersFlow.flatMapLatest { p ->
```
typotter, line 50: **Status transition gap on provider swap**

`flatMapLatest` cancels the previous inner collection and starts collecting the new provider's `status`, but any emissions during the resubscription window are missed. A downstream collector can miss the `NotReady` → `Ready` transition if it completes before the new subscription is active.

Consider seeding the new inner flow with the provider's current value:
```suggestion
            flow { emit(p.status.value); emitAll(p.status) }
```

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/OpenFeatureAPI.kt
```diff
@@ -104,18 +129,38 @@ object OpenFeatureAPI {
         }
 
         // Emit NotReady status after swapping provider
-        _statusFlow.emit(OpenFeatureStatus.NotReady)
+        // Legacy path: reset Provider/SDK-managed status.
+        if (provider !is StateManagingProvider) {
```
typotter: **NotReady emission happens outside the mutex**

For legacy providers, `_statusFlow.emit(OpenFeatureStatus.NotReady)` fires after the mutex releases, leaving a brief window where the new provider is registered but status hasn't reverted yet. Moving the emit inside the lock would make the registration and status reset atomic.

README.md
```diff
@@ -280,10 +280,19 @@ in an Android app.
...
+
+`FeatureProvider` is still fully supported. StateManagingProvider is preferred for new providers that want authoritative status ownership.
+
+Keep `status` and the flow from `observe()` aligned: when you transition between `NotReady`, `Reconciling`, and `Ready`, update `_status` and emit the matching `OpenFeatureProviderEvents` for code that listens via `OpenFeatureAPI.observe()`.
```
typotter: **Example code missing the `observe()` emission step**

The README advises keeping `status` and `observe()` aligned, but the code example doesn't show how to emit from `observe()` when `_status` transitions. The example is incomplete without that step.

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/StateManagingProvider.kt
```diff
@@ -0,0 +1,25 @@
...
+    val status: StateFlow<OpenFeatureStatus>
+
+    /**
```
typotter, line 22: Please add KDocs.

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/FeatureProvider.kt
```diff
@@ -11,8 +11,17 @@ interface FeatureProvider {
...
+     * [OpenFeatureProviderEvents.ProviderError] (on error) via
+     * [observe] before or after this function returns.
+     *
+     * The SDK waits for the first non-NotReady status derived from these events before considering
+     * initialization complete, so skipping the [observe] emission will cause [OpenFeatureAPI.setProviderAndWait]
+     * to hang indefinitely.
+     *
      * @param initialContext any initial context to be set before the provider is ready
      */
     @Throws(OpenFeatureError::class, CancellationException::class)
```
typotter, line 18: **Missing `@param` documentation on `onContextSet()`**

`onContextSet()` has a `@Throws` annotation but no `@param` docs for its parameters.

> _(Original note was at line 43; anchored to nearest covered line in this diff.)_

typotter (changes requested):
This is a tricky one full of little races. Hopefully we can find them all.
I'll have to take another look tomorrow to get a better sense of the changes from the provider's perspective

README.md
```diff
@@ -280,10 +280,19 @@ in an Android app.
 ### Develop a provider
 
 To develop a provider, you need to create a new project and include the OpenFeature SDK as a dependency.
-You’ll then need to write the provider by implementing the `FeatureProvider` interface exported by the OpenFeature SDK.
+You’ll then need to write the provider by implementing the `StateManagingProvider` interface exported by the OpenFeature SDK.
+
```
typotter, line 285: **Normative language and consistency in provider documentation**

A few issues with the new prose in the "Develop a provider" section:

1. **`StateManagingProvider` not in backticks** — inconsistent with surrounding code references.
2. **Second-person imperative** — "Keep `status` and the flow from `observe()` aligned: when you transition..." is addressed to "you" rather than specifying a contract. SDK documentation for implementors should use normative third-person.
3. **"the matching `OpenFeatureProviderEvents`" is underspecified** — does not state the required pairing (e.g. `OpenFeatureStatus.Reconciling` must be paired with `ProviderReconciling()`), nor does it explain the consequence of inconsistency.

Suggested replacement for the two prose paragraphs:

```suggestion
`FeatureProvider` is still fully supported. `StateManagingProvider` is the preferred interface for new providers that require authoritative status ownership.

The provider must keep `status` and `observe()` consistent: each time the provider transitions between `OpenFeatureStatus.NotReady`, `OpenFeatureStatus.Reconciling`, and `OpenFeatureStatus.Ready`, it must update `_status` and emit the corresponding `OpenFeatureProviderEvents` (for example, `OpenFeatureStatus.Reconciling` paired with `ProviderReconciling()`). The SDK derives `statusFlow` from `status`, and application-level handlers registered via `OpenFeatureAPI.observe()` receive the emitted events — inconsistency between the two will produce contradictory state to callers.
```

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/OpenFeatureStatus.kt
```diff
@@ -32,4 +32,9 @@ sealed interface OpenFeatureStatus {
      * The provider is reconciling its state with a context change.
      */
     object Reconciling : OpenFeatureStatus
+
+    /**
+     * The provider's configuration has changed.
+     */
+    object ConfigurationChanged : OpenFeatureStatus
```
dd-oleksii: major: this is not a valid status. It is a possible event but it [does not change provider's status.](https://openfeature.dev/specification/sections/events#requirement-535)

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/StateManagingProvider.kt
dd-oleksii, line 15: minor: this PR makes the code way to hard for me to follow and reason about, tracking in the head: (a) new and legacy code paths, (b) provider changing dynamically — possibly swapping between new and legacy code paths on the fly, (c) all of that potentially happening concurrently from multiple threads.

Given that Kotlin SDK is pre-1.0, I wonder if it makes sense to replace the old interface with the state-managing one? That's what we did [in Swift SDK](https://github.com/open-feature/swift-sdk/pull/114). That is a breaking but it makes the SDK much simpler to reason about without all `is StateManagingProvider` checks

Another option you might consider is implementing a wrapper/adapter that takes `FeatureProvider` and adapts it to `StateManagingProvider` — this way, all the legacy support is concentrated in one place and is easier to reason about (inner `provider` would be `val`, so no need to think about `OpenFeatureAPI.provider` changing from state-managing to not-state-managing or vice versa)

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/FeatureProvider.kt
```diff
@@ -11,8 +11,17 @@ interface FeatureProvider {
...
 
     /**
-     * Called by OpenFeatureAPI whenever the new Provider is registered
-     * This function should block until ready and throw exceptions if it fails to initialize
+     * Called by OpenFeatureAPI whenever the new Provider is registered.
+     *
+     * Implementations must signal their initial state by emitting
+     * [OpenFeatureProviderEvents.ProviderReady] (on success) or
+     * [OpenFeatureProviderEvents.ProviderError] (on error) via
+     * [observe] before or after this function returns.
```
dd-oleksii: minor: we should have the same requirements for `onContextSet`, right?

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/NoOpProvider.kt
```diff
@@ -1,13 +1,29 @@
...
     override suspend fun initialize(initialContext: EvaluationContext?) {
-        // no-op
+        _status.value = OpenFeatureStatus.Ready
+        events.emit(OpenFeatureProviderEvents.ProviderReady())
     }
 
     override fun shutdown() {
-        // no-op
+        _status.value = OpenFeatureStatus.NotReady
+        events.tryEmit(OpenFeatureProviderEvents.ProviderNotReady)
```
dd-oleksii: minor: `OpenFeatureProviderEvents.ProviderNotReady` is deprecated

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/OpenFeatureAPI.kt
```diff
@@ -192,15 +249,21 @@ object OpenFeatureAPI {
         } catch (e: CancellationException) {
             // This happens by design and shouldn't be treated as an error
         } catch (e: OpenFeatureError) {
-            _statusFlow.emit(OpenFeatureStatus.Error(e))
+            // Legacy path: only non-state-managing providers use _statusFlow for surfaced errors.
+            if (getProvider() !is StateManagingProvider) {
```
dd-oleksii: major(?): I believe this is not thread-safe? `getProvider()` here might return a different provider compared to the one `tryWithStatusEmitErrorHandling` was called with initially

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/NoOpProvider.kt
dd-oleksii: In Swift SDK, we're going in the opposite direction and derive status from events: see [ProviderStatusTracker](https://github.com/open-feature/swift-sdk/blob/1c206188a231462a4938c52951e97a7cb91792ac/Sources/OpenFeature/ProviderStatusTracker.swift). Mostly because events are more rich than statuses

README.md
maxnrp: Actually, my bad here. This issue’s specs clearly state that ‘legacy behavior is deprecated and will be removed in the next major version.’ Given that, we should discourage implementing it. It shouldn’t be opt-in, but rather compatible for now.

I left a note on that at the end of the ‘Develop a provider’ section so we are not mixing this with the new instructions.

kotlin-sdk/api/android/kotlin-sdk.api
maxnrp: I think the method here [adds itself](https://medium.com/@AlexanderObregon/kotlin-property-getters-and-setters-behind-the-scenes-de156fcbe086), but the need for the property status, yes, it is part of the [spec](https://github.com/open-feature/spec/issues/365#issuecomment-4046615584).

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/StateManagingProvider.kt
maxnrp: I decided to leave FeatureProvider unchanged and redeclare here the functions whose contracts changed in the new interface, with their specific requirements.

README.md
maxnrp: Here, I believe only .shutdown() wasn't aligned.
All other SMP-relevant methods in the example alter the feature provider status and emit it to events, which is returned by the observe() flow.

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/OpenFeatureAPI.kt
maxnrp: yes, this is a consistency issue. I've removed the `StateManagingProvider` evaluation here because it only matters from where it is being called (only legacy paths due to event emission).

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/FeatureProvider.kt
maxnrp: I've restored _FeatureProvider.kt_ because I think it shouldn't be touched if we are keeping it as legacy for future removal.

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/StateManagingProvider.kt
maxnrp: > Given that Kotlin SDK is pre-1.0, I wonder if it makes sense to replace the old interface with the state-managing one?

... that would be simple.

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/NoOpProvider.kt
maxnrp: true, I've noticed JS SDK also has a statusTracker. But that's it, having one source of truth seems to be what matters the most.

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/StateManagingProvider.kt
maxnrp: About the adaptor, Java and JS PoC was made using an interfaces, so that was my lead, but an adaptor seems indeed like a cleaner option, but would it be worthy if FeatureProvider will disappear?

maxnrp:
Let's close this for now... and see where we go with the legacy provider adaptor.

PR #231, opened 2026-04-13
feat: add Domains (formerly Clients) (#185)

### LINKED ISSUE #185: Missing domain support
https://openfeature.dev/specification/sections/flag-evaluation/#requirement-113

mstepien:
## Description
This PR introduces the **Domains** (formerly Named Clients) feature to the OpenFeature Kotlin SDK, ensuring full compliance with the OpenFeature Domains specification. 

Domains allow logical binding of clients with specific providers, enabling the use of multiple providers in a single application. If no specific provider or context is bound to a given domain, the system gracefully falls back to the globally configured default provider and evaluation context. 

## Changes Made
- **Client Metadata Updates**: Added `domain: String?` to `ClientMetadata`. Concurrently deprecated `name` in favor of `domain` to align with the wider OpenFeature schema and specification.
- **Support for Domain-Bound Providers**: Augmented `OpenFeatureAPI` with `domain`-parameterized methods (e.g., `setProvider(domain, provider)`, `getProvider(domain)`, `getProviderStatus(domain)`).
- **Domain State Isolation**: Introduced `ProviderRepository` and `DomainState` to isolate lifecycle, observers, events, and execution flow tracking per domain in a thread-safe manner using atomic flow substitutions and `Mutex`.
- **Domain-Specific Evaluation Context**: Permitted setting and retrieving `EvaluationContext` on a per-domain basis `setEvaluationContext(domain, context)`. A domain client will first attempt to use its specific contextual state before defaulting back to the global execution context.
- **Event Flow Isolation**: Ensured that hot-swapping providers and evaluating events dynamically bubbles up isolated state and `OpenFeatureStatus` to the individual domain correctly.
- **E2E Validation**: Provided test coverage in `ProviderRepositoryTest` and `DomainE2ETest` to enforce atomic state concurrency, dynamically mapped fallback lifecycles, and strict hook/event propagation rules bounded to isolated domains.

## Related Issues
Closes #185

## Testing
- [x] End-to-end domain evaluation test assertions executed (`DomainE2ETest`)
- [x] Concurrency and flow-mapping assertions completed (`ProviderRepositoryTest`)
- [x] Evaluated hook propagation order locally.
- [x] Unit/Integration tests all passed successfully on JVM target.

## Reviewer Focus
- Please verify the fallback logic in `OpenFeatureAPI` where missing domain contexts gracefully fallback to global contexts without breaking isolation.
- Sanity-check the coroutine scope management and flow-switching inside `ProviderRepository` (`getStateFlow()`) for potential leaks when updating providers rapidly.

### DIFF
--- a/kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/OpenFeatureAPI.kt
+++ b/kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/OpenFeatureAPI.kt
@@ -131,6 +131,8 @@ object OpenFeatureAPI {
         initialContext: EvaluationContext? = null,
         isGlobalContext: Boolean = false
     ) {
+        repository.attachProvider(provider)
+
         // Atomically swap the old and new provider to prevent race conditions
         val oldProvider = state.providerMutex.withLock {
             val current = state.provider
@@ -139,15 +141,17 @@ object OpenFeatureAPI {
             current
         }
 
-        if (initialContext != null) {
-            updateContextOnProviderSet(state, initialContext, isGlobalContext, dispatcher)
+        // Shutdown the previous provider isolated from stream errors
+        if (repository.detachProvider(oldProvider)) {
+            try {
+                oldProvider.shutdown()
+            } catch (e: Exception) {
+                // Ignore termination exceptions from dead configurations natively securely
+            }
         }
 
-        // Shutdown the previous provider isolated from stream errors
-        try {
-            oldProvider.shutdown()
-        } catch (e: Exception) {
-            // Ignore termination exceptions from dead configurations natively securely
+        if (initialContext != null) {
+            updateContextOnProviderSet(state, initialContext, isGlobalContext, dispatcher)
         }
 
         initializeProvider(state, provider, dispatcher)
--- a/kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/OpenFeatureAPI.kt
+++ b/kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/OpenFeatureAPI.kt
@@ -60,7 +60,7 @@ object OpenFeatureAPI {
     /**
      * Set the [FeatureProvider] for a specific domain.
      */
-    fun setProvider(
+    internal fun setProvider(
         domain: String?,
         provider: FeatureProvider,
         dispatcher: CoroutineDispatcher = Dispatchers.Default,
@@ -108,7 +108,7 @@ object OpenFeatureAPI {
     /**
      * Set the [FeatureProvider] for a specific domain. This method will block until the provider is initialized.
      */
-    suspend fun setProviderAndWait(
+    internal suspend fun setProviderAndWait(
         domain: String?,
         provider: FeatureProvider,
         initialContext: EvaluationContext? = null,
--- a/kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/OpenFeatureAPI.kt
+++ b/kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/OpenFeatureAPI.kt
@@ -10,7 +10,6 @@ import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.SupervisorJob
 import kotlinx.coroutines.flow.Flow
-import kotlinx.coroutines.flow.catch
 import kotlinx.coroutines.flow.distinctUntilChanged
 import kotlinx.coroutines.flow.filterIsInstance
 import kotlinx.coroutines.flow.flatMapLatest
@@ -204,6 +203,7 @@ object OpenFeatureAPI {
                 val resolvedContext = state.contextMutex.withLock { state.mergedContext } ?: globalCtx
                 provider.initialize(resolvedContext)
                 state.emitStatus(OpenFeatureStatus.Ready)
+                state.emitEvent(OpenFeatureProviderEvents.ProviderReady())
             }
         }
     }
@@ -431,9 +431,11 @@ object OpenFeatureAPI {
 
             activeProvider?.let { provider ->
                 state.emitStatus(OpenFeatureStatus.Reconciling)
+                state.emitEvent(OpenFeatureProviderEvents.ProviderReconciling())
                 tryWithStatusEmitErrorHandling(state) {
                     provider.onContextSet(oldContext, newContext)
                     state.emitStatus(OpenFeatureStatus.Ready)
+                    state.emitEvent(OpenFeatureProviderEvents.ProviderContextChanged())
                 }
             }
         }
@@ -446,11 +448,22 @@ object OpenFeatureAPI {
             // This happens by design and shouldn't be treated as an error
         } catch (e: OpenFeatureError) {
             state.emitStatus(OpenFeatureStatus.Error(e))
+            state.emitEvent(
+                OpenFeatureProviderEvents.ProviderError(
+                    eventDetails = OpenFeatureProviderEvents.EventDetails(
+                        message = e.message ?: "Unknown error",
+                        errorCode = e.errorCode()
+                    )
+                )
+            )
         } catch (e: Throwable) {
-            state.emitStatus(
-                OpenFeatureStatus.Error(
-                    OpenFeatureError.GeneralError(
-                        e.message ?: "Unknown error"
+            val generalError = OpenFeatureError.GeneralError(e.message ?: "Unknown error")
+            state.emitStatus(OpenFeatureStatus.Error(generalError))
+            state.emitEvent(
+                OpenFeatureProviderEvents.ProviderError(
+                    eventDetails = OpenFeatureProviderEvents.EventDetails(
+                        message = generalError.message,
+                        errorCode = generalError.errorCode()
                     )
                 )
             )
@@ -544,43 +557,22 @@ object OpenFeatureAPI {
     internal fun getProvidersFlowForDomain(domain: String?): Flow<FeatureProvider> =
         repository.getStateFlow(domain).flatMapLatest { it.providersFlow }
 
+    @PublishedApi
+    @OptIn(ExperimentalCoroutinesApi::class)
+    internal fun getEventsFlowForDomain(domain: String?): Flow<OpenFeatureProviderEvents> =
+        repository.getStateFlow(domain).flatMapLatest { it.eventsFlow }
+
     /**
      * Observe events from currently configured default provider.
      */
     @OptIn(ExperimentalCoroutinesApi::class)
     inline fun <reified T : OpenFeatureProviderEvents> observe(): Flow<T> =
-        getProvidersFlowForDomain(null)
-            .flatMapLatest { provider ->
-                provider.observe().catch { cause ->
-                    emit(
-                        OpenFeatureProviderEvents.ProviderError(
-                            OpenFeatureProviderEvents.EventDetails(
-                                message = cause.message ?: "Provider observe() crashed",
-                                errorCode = dev.openfeature.kotlin.sdk.exceptions.ErrorCode.GENERAL
-                            )
-                        )
-                    )
-                }
-            }
-            .filterIsInstance()
+        getEventsFlowForDomain(null).filterIsInstance()
 
     /**
      * Observe events from currently configured provider for the specified domain.
      */
     @OptIn(ExperimentalCoroutinesApi::class)
     inline fun <reified T : OpenFeatureProviderEvents> observe(domain: String?): Flow<T> =
-        getProvidersFlowForDomain(domain)
-            .flatMapLatest { provider ->
-                provider.observe().catch { cause ->
-                    emit(
-                        OpenFeatureProviderEvents.ProviderError(
-                            OpenFeatureProviderEvents.EventDetails(
-                                message = cause.message ?: "Provider observe() crashed",
-                                errorCode = dev.openfeature.kotlin.sdk.exceptions.ErrorCode.GENERAL
-                            )
-                        )
-                    )
-                }
-            }
-            .filterIsInstance()
+        getEventsFlowForDomain(domain).filterIsInstance()
 }
--- a/kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/OpenFeatureAPI.kt
+++ b/kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/OpenFeatureAPI.kt
@@ -205,9 +205,38 @@ object OpenFeatureAPI {
             tryWithStatusEmitErrorHandling(state) {
                 val globalCtx = globalContextMutex.withLock { context }
                 val resolvedContext = state.contextMutex.withLock { state.mergedContext } ?: globalCtx
-                provider.initialize(resolvedContext)
-                state.emitStatus(OpenFeatureStatus.Ready)
-                state.emitEvent(OpenFeatureProviderEvents.ProviderReady())
+
+                if (provider !is NoOpProvider) {
+                    val initMutex = repository.getInitMutex(provider)
+                    initMutex.withLock {
+                        if (!repository.isInitialized(provider)) {
+                            provider.initialize(resolvedContext)
+                            repository.markInitialized(provider)
+                            repository.updateGlobalProviderStatus(provider, OpenFeatureStatus.Ready)
+                            state.emitStatus(OpenFeatureStatus.Ready)
+                            state.emitEvent(OpenFeatureProviderEvents.ProviderReady())
+                        } else {
+                            val currentStatus = repository.getGlobalProviderStatus(provider) ?: OpenFeatureStatus.Ready
+                            state.emitStatus(currentStatus)
+                            when (currentStatus) {
+                                is OpenFeatureStatus.Ready -> state.emitEvent(OpenFeatureProviderEvents.ProviderReady())
+                                is OpenFeatureStatus.Error -> state.emitEvent(
+                                    OpenFeatureProviderEvents.ProviderError(
+                                        dev.openfeature.kotlin.sdk.events.OpenFeatureProviderEvents.EventDetails(
+                                            message = "Provider in error state"
+                                        )
+                                    )
+                                )
+                                is OpenFeatureStatus.Stale -> state.emitEvent(OpenFeatureProviderEvents.ProviderStale())
+                                else -> {}
+                            }
+                        }
+                    }
+                } else {
+                    provider.initialize(resolvedContext)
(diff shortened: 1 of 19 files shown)


kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/OpenFeatureAPI.kt
```diff
@@ -205,25 +296,39 @@ object OpenFeatureAPI {
...
     fun getEvaluationContext(): EvaluationContext? {
         return context
     }
 
     /**
-     * Get the [ProviderMetadata] for the current [FeatureProvider].
+     * Get the [EvaluationContext] for the specified domain. If not set, returns the global context.
+     */
+    fun getEvaluationContext(domain: String?): EvaluationContext? {
+        return repository.getState(domain).context ?: context
```
typotter: Section 3.2.3 of the spec mandates that the context be merged. See: https://openfeature.dev/specification/sections/evaluation-context#requirement-323

`getEvaluationContext(domain)` currently uses `state.context ?: context` — if domain context is set, all global context attributes are silently discarded. The spec requires merging: global context as the base, with domain context keys taking higher precedence.

`ImmutableContext` has no `merge()` method, so this can't be implemented without adding one. Suggested approach:

```kotlin
fun EvaluationContext.mergeWith(higher: EvaluationContext): EvaluationContext {
    val merged = asMap().toMutableMap()
    merged.putAll(higher.asMap())
    val targetingKey = higher.getTargetingKey().ifEmpty { getTargetingKey() }
    return ImmutableContext(targetingKey, merged)
}
```

Then replace line 309 with:
```kotlin
val domainCtx = repository.getState(domain).context
return when {
    domainCtx == null -> context
    context == null -> domainCtx
    else -> context.mergeWith(domainCtx)
}
```

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/ProviderRepository.kt
```diff
@@ -0,0 +1,163 @@
...
+import kotlinx.coroutines.flow.map
+import kotlinx.coroutines.flow.onEach
+import kotlinx.coroutines.flow.retryWhen
+import kotlinx.coroutines.flow.update
+import kotlinx.coroutines.sync.Mutex
+import kotlinx.coroutines.sync.withLock
+import kotlinx.coroutines.withContext
+import kotlin.concurrent.Volatile
+
+class DomainState {
```
typotter: `DomainState` and `ProviderRepository` should be `internal`/package-private and not part of the public API. Remove `getRepository()` from `OpenFeatureAPI` or mark it `internal`.

The binary API file confirms these are currently public: `DomainState` exposes `get_statusFlow(): MutableSharedFlow`, `getProviderMutex()`, `getSetProviderJob()` — internal coroutine infrastructure that consumers should never access directly.

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/ProviderRepository.kt
```diff
@@ -0,0 +1,163 @@
...
+
+        return domainsFlow.value[domain] ?: repositoryMutex.withLock {
+            domainsFlow.value[domain] ?: DomainState().also { newState ->
+                domainsFlow.update { currentMap -> currentMap + (domain to newState) }
+            }
+        }
+    }
+
+    fun getStateFlow(domain: String?): Flow<DomainState> {
+        if (domain == null) return MutableStateFlow(defaultDomainState)
```
typotter: The null domain state flow is recreated on every get, so no subscribers will ever get events.

Line 142 constructs a `new MutableStateFlow(defaultDomainState)` on each call — each subscriber gets an independent static snapshot. Compare with the domain-specific path at line 143 which correctly uses `domainsFlow.map { ... }` (reactive).

Fix: memoize as a field.
```kotlin
private val defaultStateFlow = MutableStateFlow(defaultDomainState)

fun getStateFlow(domain: String?): Flow<DomainState> {
    if (domain == null) return defaultStateFlow
    return domainsFlow.map { it[domain] ?: defaultDomainState }.distinctUntilChanged()
}
```

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/ProviderRepository.kt
```diff
@@ -0,0 +1,163 @@
...
+
+    suspend fun shutdown() {
+        setProviderJob?.cancel(CancellationException("Provider set job was cancelled due to shutdown"))
+        setEvaluationContextJob?.cancel(CancellationException("Set context job was cancelled due to shutdown"))
+        providerMutex.withLock {
+            domainScope?.cancel(CancellationException("DomainScope was cancelled due to shutdown"))
+            domainScope = null
+            currentDispatcher = null
+        }
+        provider.shutdown()
```
typotter: The reset and shutdown should be in the same mutex block to prevent the provider reference from being swapped after the lock is released.

`provider.shutdown()` at line 115 reads `providersFlow.value` (via the computed property) after the `providerMutex` was released at line 114. A concurrent `setProviderInternal()` can swap the provider in that window, causing the newly-installed provider to be shut down instead.

```kotlin
suspend fun shutdown() {
    setProviderJob?.cancel(...)
    setEvaluationContextJob?.cancel(...)
    val providerToShutdown: FeatureProvider
    providerMutex.withLock {
        domainScope?.cancel(...)
        domainScope = null
        currentDispatcher = null
        providerToShutdown = providersFlow.value
        providersFlow.value = NoOpProvider()
    }
    providerToShutdown.shutdown()  // called after lock release, but on captured reference
    _statusFlow.tryEmit(OpenFeatureStatus.NotReady)
}
```

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/OpenFeatureAPI.kt
typotter: Section 3.2.3 of the spec mandates that the context be merged. See: https://openfeature.dev/specification/sections/evaluation-context#requirement-323

`getEvaluationContext(domain)` currently uses `state.context ?: context` — if domain context is set, all global context attributes are silently discarded. The spec requires merging: global context as the base, with domain context keys taking higher precedence.

`ImmutableContext` has no `merge()` method, so this can't be implemented without adding one. Suggested approach:

```kotlin
fun EvaluationContext.mergeWith(higher: EvaluationContext): EvaluationContext {
    val merged = asMap().toMutableMap()
    merged.putAll(higher.asMap())
    val targetingKey = higher.getTargetingKey().ifEmpty { getTargetingKey() }
    return ImmutableContext(targetingKey, merged)
}
```

Then replace line 309 with:
```kotlin
val domainCtx = repository.getState(domain).context
return when {
    domainCtx == null -> context
    context == null -> domainCtx
    else -> context.mergeWith(domainCtx)
}
```

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/ProviderRepository.kt
typotter: `DomainState` and `ProviderRepository` should be `internal` and not part of the public API. Remove `getRepository()` from `OpenFeatureAPI` or mark it `internal`.

The binary API file confirms these are currently public: `DomainState` exposes `get_statusFlow(): MutableSharedFlow`, `getProviderMutex()`, `getSetProviderJob()` — internal coroutine infrastructure that consumers should never access directly.

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/ProviderRepository.kt
typotter: The null domain state flow is recreated on every get, so no subscribers will ever get events.

Line 142 constructs a `new MutableStateFlow(defaultDomainState)` on each call — each subscriber gets an independent static snapshot. Compare with the domain-specific path at line 143 which correctly uses `domainsFlow.map { ... }` (reactive).

Fix: memoize as a field.
```kotlin
private val defaultStateFlow = MutableStateFlow(defaultDomainState)

fun getStateFlow(domain: String?): Flow<DomainState> {
    if (domain == null) return defaultStateFlow
    return domainsFlow.map { it[domain] ?: defaultDomainState }.distinctUntilChanged()
}
```

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/ProviderRepository.kt
typotter: The reset and shutdown should be in the same mutex block to prevent the provider reference from being swapped after the lock is released.

`provider.shutdown()` at line 115 reads `providersFlow.value` after `providerMutex` is released at line 114. A concurrent `setProviderInternal()` can swap the provider in that window, causing the newly-installed provider to be shut down instead.

```kotlin
suspend fun shutdown() {
    setProviderJob?.cancel(...)
    setEvaluationContextJob?.cancel(...)
    val providerToShutdown: FeatureProvider
    providerMutex.withLock {
        domainScope?.cancel(...)
        domainScope = null
        currentDispatcher = null
        providerToShutdown = providersFlow.value
        providersFlow.value = NoOpProvider()
    }
    providerToShutdown.shutdown()
    _statusFlow.tryEmit(OpenFeatureStatus.NotReady)
}
```

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/ProviderRepository.kt
typotter: `DomainState` and `ProviderRepository` should be `internal` and not part of the public API. Remove `getRepository()` from `OpenFeatureAPI` or mark it `internal`.

The binary API file confirms these are currently public: `DomainState` exposes `get_statusFlow(): MutableSharedFlow`, `getProviderMutex()`, `getSetProviderJob()` — internal coroutine infrastructure that consumers should never access directly.

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/ProviderRepository.kt
typotter: The null domain state flow is recreated on every get, so no subscribers will ever get events.

`MutableStateFlow(defaultDomainState)` is constructed on each call — each subscriber gets an independent static snapshot. Compare with the domain-specific path at line 143 which correctly uses `domainsFlow.map { ... }` (reactive).

Fix: memoize as a field.
```kotlin
private val defaultStateFlow = MutableStateFlow(defaultDomainState)

fun getStateFlow(domain: String?): Flow<DomainState> {
    if (domain == null) return defaultStateFlow
    return domainsFlow.map { it[domain] ?: defaultDomainState }.distinctUntilChanged()
}
```

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/ProviderRepository.kt
typotter: The reset and shutdown should be in the same mutex block to prevent the provider reference from being swapped after the lock is released.

`provider.shutdown()` at line 115 reads `providersFlow.value` after `providerMutex` is released at line 114. A concurrent `setProviderInternal()` can swap the provider in that window, causing the newly-installed provider to be shut down instead.

```kotlin
suspend fun shutdown() {
    setProviderJob?.cancel(...)
    setEvaluationContextJob?.cancel(...)
    val providerToShutdown: FeatureProvider
    providerMutex.withLock {
        domainScope?.cancel(...)
        domainScope = null
        currentDispatcher = null
        providerToShutdown = providersFlow.value
        providersFlow.value = NoOpProvider()
    }
    providerToShutdown.shutdown()
    _statusFlow.tryEmit(OpenFeatureStatus.NotReady)
}
```

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/OpenFeatureAPI.kt
typotter: Section 3.2.3 of the spec mandates that the context be merged. See: https://openfeature.dev/specification/sections/evaluation-context#requirement-323

`getEvaluationContext(domain)` currently uses `state.context ?: context` — if domain context is set, all global context attributes are silently discarded. The spec requires merging: global context as the base, with domain context keys taking higher precedence.

`ImmutableContext` has no `merge()` method, so this requires adding one. Suggested approach:

```kotlin
fun EvaluationContext.mergeWith(higher: EvaluationContext): EvaluationContext {
    val merged = asMap().toMutableMap()
    merged.putAll(higher.asMap())
    val targetingKey = higher.getTargetingKey().ifEmpty { getTargetingKey() }
    return ImmutableContext(targetingKey, merged)
}
```

Then replace line 309 with:
```kotlin
val domainCtx = repository.getState(domain).context
return when {
    domainCtx == null -> context
    context == null -> domainCtx
    else -> context.mergeWith(domainCtx)
}
```

typotter:
Spec 3.2.2.4 requires the API to have a mechanism to manage (create and remove) evaluation context for an associated domain.

There is no `clearEvaluationContext(domain)` method. Once domain context is set, it can be replaced but not cleared to revert to global fallback. Worth considering whether `setEvaluationContext(domain, emptyContext)` is sufficient or whether an explicit clear is needed.

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/ProviderRepository.kt
mstepien: Fixed with https://github.com/open-feature/kotlin-sdk/pull/231/changes/bba9540d4ec3b9c51d3c7dd7a3181780c69261d0

mstepien:
> Spec 3.2.2.4 requires the API to have a mechanism to manage (create and remove) evaluation context for an associated domain.
> 
> There is no `clearEvaluationContext(domain)` method. Once domain context is set, it can be replaced but not cleared to revert to global fallback. Worth considering whether `setEvaluationContext(domain, emptyContext)` is sufficient or whether an explicit clear is needed.

fixed with https://github.com/open-feature/kotlin-sdk/pull/231/changes/ca14dfa7f8b48fc484d41e0408d3fbbb72f1ea2a

mstepien:
### 2026-04-22-23 PR Update: Multi-Domain Isolation, Lifecycle Management, and Specification Compliance

This update brings core specification's requirements regarding isolated eventing, shared provider lifecycles, and evaluation context management.

#### Provider Lifecycle & Deduplication
* **Thread-safe Shared Provider Tracking (Spec 1.1.2.3):** Implemented global reference counting in `ProviderRepository`. `OpenFeatureAPI` now ensures a shared provider's `shutdown()` function is *only* invoked once the provider's global reference count across all domains safely reaches zero.
* **Initialization Deduplication (Spec 1.1.2.2):** Introduced `Mutex`-backed initialization tracking to guarantee that a provider bound to multiple domains is only actively initialized exactly once.
* **Global Health Ledger (Spec 1.1.2.2):** Created a centralized status tracker that securely syncs background health transitions (like `ProviderError` crashes or `ProviderStale`) across all active domains sharing a provider instance.
* **Synthetic Event Emission:** Ensures that domains successfully bypassing redundant initialization actively poll the global ledger and synthetically rebroadcast the synchronized state (e.g., `ProviderReady`) to their downstream clients.

#### Domain-Isolated Eventing 
* **Client Event Observability (Specs 5.2.1, 5.2.7):** Added `observeEvents()` and `observe<T>()` APIs to the `Client` interface, natively routing domain-isolated provider events directly to end users.
* **Context Event Generation (Specs 5.3.4.1-3):** Hardened the `setEvaluationContext` workflow to securely isolate event emissions. It now emits `ProviderReconciling` prior to execution, `ProviderContextChanged` on success, and catches unhandled exceptions to emit `ProviderError` if a crash occurs within the provider's `onContextSet`.

#### Evaluation Context Management
* **Domain Context Merging (Spec 3.2.3):** Added context merging capabilities bound per domain, including an internal cache to eliminate garbage collection overhead during rapid, highly-concurrent flag evaluations.
* **Context Teardown (Spec 3.2.2.4):** Added the ability to clear and manage evaluation contexts natively per domain, fortified with granular `contextMutex` locks to prevent TOCTOU race condition deadlocks.

#### Architecture Hardening & Tech Debt
* **State Encapsulation:** Refactored architecture to hide internal state objects (`DomainState`) from the public API.
* **Memory & Thread Safety:** Memoized the `defaultDomainState` flow to prevent allocations, eliminated race conditions during rapid provider swapping, and ensured all phantom background observer coroutines cancel out completely upon `OpenFeatureAPI.shutdown()`.

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/OpenFeatureAPI.kt
```diff
@@ -23,126 +19,275 @@ import kotlinx.coroutines.sync.withLock
...
-        setProviderJob?.cancel(CancellationException("Provider set job was cancelled due to new provider"))
-        this.setProviderJob = CoroutineScope(SupervisorJob() + dispatcher).launch {
-            setProviderInternal(provider, dispatcher, initialContext)
+        launchProviderJob(null, provider, dispatcher, initialContext, isGlobalContext = true)
+    }
+
+    /**
+     * Set the [FeatureProvider] for a specific domain.
+     */
+    fun setProvider(
```
typotter: **API surface — domain-parameterized `setProvider` on OpenFeatureAPI**

Every other OpenFeature SDK that supports domains (JS, Java, .NET, Python, Go) exposes `setProvider(domain, provider)` on the API singleton, but this creates an ergonomic problem: the caller must hold onto the domain string and pass it again at `setProvider` time, even though the domain binding was already established at `getClient(domain)` time.

Preferred pattern (consistent with how the JS/Java SDKs handle `setContext` at the client level):

```kotlin
// entry point: domain only appears here
val client = OpenFeatureAPI.getClient("checkout")

// domain-scoped operations on the client — no domain string needed again
client.setProvider(myProvider)
client.setProviderAndWait(myProvider)
```

This means `setProvider` / `setProviderAndWait` with a `domain: String?` parameter should not exist on `OpenFeatureAPI`. They should be methods on the `Client` interface instead. The `client` object already encapsulates the domain — there is no reason for callers to re-supply it.

This is the only API surface change needed to make the Kotlin SDK congruent with the design principle that `getClient(domain)` is the sole place the domain string is required.

mstepien: refactored with 16b88f1

PR #232, opened 2026-04-13
feat: Merge eval context if provided

PenguinDan:
## This PR
EvalContext can now be provided from evaluation functions which allows us to merge the eval context passed during flag evaluation, merged with the global context

Following https://github.com/open-feature/java-sdk/blob/main/src/main/java/dev/openfeature/sdk/OpenFeatureClient.java

### How to test
Unit tests created


kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/Features.kt
```diff
@@ -3,37 +3,103 @@ package dev.openfeature.kotlin.sdk
 interface Features {
     fun getBooleanValue(key: String, defaultValue: Boolean): Boolean
     fun getBooleanValue(key: String, defaultValue: Boolean, options: FlagEvaluationOptions): Boolean
+    fun getBooleanValue(
+        key: String,
+        defaultValue: Boolean,
+        options: FlagEvaluationOptions = FlagEvaluationOptions(),
+        evaluationContext: EvaluationContext? = null,
+    ): Boolean
```
PenguinDan, line 11: Seems like it would be a breaking change to combine all of these with default params

Happy to merge these if we are okay with that, I do think that would be a simpler api and I do see a breaking change that seems like will be released together with this potentially

typotter:
I think this would be counter to the [Client SDK spec](https://openfeature.dev/specification/sections/evaluation-context#conditional-requirement-3222)
What's the use case ?

PR #234, opened 2026-04-15
feat!: Add observe() method to Client interface

maxnrp:
### Intent
Add a client-level way to observe provider events (`OpenFeatureProviderEvents`) with the same behavior as `OpenFeatureAPI.observe()`. https://github.com/orgs/open-feature/projects/29/views/2?pane=issue&itemId=149893842&issue=open-feature%7Ckotlin-sdk%7C203

### Motivation
`OpenFeatureAPI` already exposes observe<T>() over the current provider’s event stream.
Client exposed `statusFlow` (`OpenFeatureStatus`) but had no analogue for typed provider events.

### Changes
**Client**
`observe(): Flow<OpenFeatureProviderEvents>` — Cold flow of events from the SDK’s current FeatureProvider, aligned with `FeatureProvider.observe() `/ the same pipeline as `OpenFeatureAPI.observe()`.

`inline fun <reified T : OpenFeatureProviderEvents> Client.observe(): Flow<T>` — extension that filters with `filterIsInstance() `mirroring `OpenFeatureAPI.observe())` [pattern](https://github.com/open-feature/kotlin-sdk/blob/2bd38d1ea11098f11b0c2d78d23ab1fb5fa864d6/kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/OpenFeatureAPI.kt#L266-L268).

**OpenFeatureClient**
Implements `observe()` by streaming from `openFeatureAPI.observe<OpenFeatureProviderEvents>() `so behavior stays identical to the API-level observe path — to be further narrowed by type based on author need.

### Tests
`clientObserveMatchesApiObserveWhenCollectingAllProviderEvents` — collects OpenFeatureAPI.observe<OpenFeatureProviderEvents>() and client.observe<OpenFeatureProviderEvents>() in parallel over setProviderAndWait / shutdown, then asserts equal event lists (parity check).

`clientObserveFiltersByReifiedEventType` — collects evens from a particular assigned event type.

### Breaking changes

JVM/Android binary API.


kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/Client.kt
```diff
@@ -1,11 +1,28 @@
...
+
     fun addHooks(hooks: List<Hook<*>>)
-}
\ No newline at end of file
+}
+
+/**
+ * Observe provider events of type [T], matching [OpenFeatureAPI.observe].
+ */
+@OptIn(ExperimentalCoroutinesApi::class)
```
mstepien: Minor: Redundant `@OptIn` Annotation.
Neither the `Client.observe()` interface method nor `kotlinx.coroutines.flow.filterIsInstance` are marked as experimental

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/OpenFeatureClient.kt
```diff
@@ -26,6 +29,10 @@ class OpenFeatureClient(
 
     override val statusFlow = openFeatureAPI.statusFlow
 
+    @OptIn(ExperimentalCoroutinesApi::class)
+    override fun observe(): Flow<OpenFeatureProviderEvents> =
+        openFeatureAPI.observe<OpenFeatureProviderEvents>()
```
mstepien, line 35: **Medium: Risk of Unintentional Redundant Provider Subscriptions**

Every time `.collect` is called on the flow returned by `client.observe()`, it triggers an entirely **new subscription** via `flatMapLatest` into the underlying `provider.observe()`. If multiple clients are observing events (e.g., UI components), the SDK will make numerous potentially expensive and redundant registration calls to the current `FeatureProvider`. 


Instead of re-evaluating the flow per-caller, consider cache and distribute events centrally using a SharedFlow.

mstepien (changes requested):
Found 1 minor and 1 medium potential issues to address.

kotlin-sdk/src/commonTest/kotlin/dev/openfeature/kotlin/sdk/ProviderEventingTests.kt
```diff
@@ -142,4 +142,59 @@ class ProviderEventingTests {
...
+            client.observe<OpenFeatureProviderEvents>().collect { fromClient.add(it) }
+        }
+
+        OpenFeatureAPI.setProviderAndWait(provider, initialContext = ImmutableContext("ctx"))
+        testScheduler.advanceUntilIdle()
+        OpenFeatureAPI.shutdown()
+        apiJob.cancelAndJoin()
+        clientJob.cancelAndJoin()
+
+        assertEquals(fromApi, fromClient)
```
typotter, line 170: Please add `assertTrue(fromApi.isNotEmpty())` to ensure that the sets aren't empty.

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/Client.kt
```diff
@@ -1,11 +1,28 @@
 package dev.openfeature.kotlin.sdk
 
+import dev.openfeature.kotlin.sdk.events.OpenFeatureProviderEvents
+import kotlinx.coroutines.ExperimentalCoroutinesApi
```
typotter: `import kotlinx.coroutines.ExperimentalCoroutinesApi` unnecessary as `@OptIn` is removed

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/OpenFeatureClient.kt
maxnrp: I see, but is it that much of a burden? The solution look so noisy. I could maybe set a shared coroutine for each client's flow, but is this a real contribute? Because it would still be 1 different stream per client, not 1 shared by them all (a getClient limitation, based on what this get returns), and what about clients leaving non-cancelled CoroutineScope open?

I would like to have some opinion here.

nicklasl: My take on this (but tbh I haven't given it much thought):
I can see how we can take some pain off provider authors and handle it but if that's the case we should fix this at the `OpenFeatureAPI.observe()` level, or not at all. Adding a fix only to this part creates a confusing asymmetry.
Suggesting we accept this and open a separate issue for "centralize event distribution in `OpenFeatureAPI.observe()`".

nicklasl (approved):
LGTM.
I guess one can argue about if we should consider this breaking or not, I don't intent for users to implement their own `Client` but it might very well be overridden in a test suite so there will be some pain (which speaks to that it is correctly marked as breaking).

I am not in favour of doing a default implementation.

typotter:
Agree with fixing upstream. filed https://github.com/open-feature/kotlin-sdk/issues/245

typotter (approved):

PR #235, opened 2026-04-15
docs: fix inaccuracies in README code examples

jonathannorris:
## Summary

- Fix multiple code examples in the README that would not compile as written

## Changes

1. **Quick start**: `default = false` -> `defaultValue = false` (actual named parameter in `Features.kt`)
2. **Eventing**: Added required type parameter to `OpenFeatureAPI.observe<OpenFeatureProviderEvents>()` (reified generic function)
3. **Eventing**: Fixed `setProviderAndWait` parameter order from `(provider, dispatcher, context)` to `(provider, context, dispatcher)` to match actual signature
4. **Shutdown**: Wrapped `OpenFeatureAPI.shutdown()` in a coroutine scope since it is a `suspend` function
5. **Provider example**: `Metadata` -> `ProviderMetadata` (actual type in `FeatureProvider` interface)

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,7 +104,7 @@ coroutineScope.launch(Dispatchers.Default) {
   val client = OpenFeatureAPI.getClient()
 
   // get a bool flag value
-  client.getBooleanValue("boolFlag", default = false)
+  client.getBooleanValue("boolFlag", defaultValue = false)
 }
 ```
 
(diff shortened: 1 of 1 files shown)


typotter (approved):
All five fixes verified against the actual API signatures — correct.

nicklasl (approved):

PR #238, opened 2026-04-17
feat!: telemetry utils (#206)

### LINKED ISSUE #206: Add Telemetry Utils
**Specification References:**
- [Appendix A](https://openfeature.dev/specification/appendix-a): Telemetry hook patterns.

Consider adding a telemetry utils, similar to:

- https://github.com/open-feature/js-sdk/blob/f3f368586e3cb282e84f98a066d975a2c22d3b4c/packages/shared/src/telemetry/attributes.ts#L5
- https://github.com/open-feature/js-sdk/blob/f3f368586e3cb282e84f98a066d975a2c22d3b4c/packages/shared/src/telemetry/evaluation-event.ts#L3

These make it very easy to export telemetry data in custom hooks.

mstepien:
Signed-off-by: Marcin Stepien <marcin.stepien@fluxon.com>

## [Add Telemetry Utils #206](https://github.com/open-feature/kotlin-sdk/issues/206)
Adds Appendix D (Observability/Telemetry) mapping utilities to the Kotlin SDK. 
- Adds `dev.openfeature.kotlin.sdk.telemetry.EvaluationEvent` data construct mapped to OpenTelemetry 
- Adds `dev.openfeature.kotlin.sdk.telemetry.Telemetry` static mapping object exposing the `createEvaluationEvent()` bridging utility.
- Implements parsing logic translating `HookContext` and `FlagEvaluationDetails` payloads, accounting for Error state mutations, Context evaluations, and natively unwrapping SDK `Value` objects iteratively down to pure primitives/Maps.
- Adds unit and integration testing
### Related Issues
Fixes #206 

### Breaking changes
`EvaluationEvent` and `createEvaluationEvent()` moved from `dev.openfeature.kotlin.sdk `to `dev.openfeature.kotlin.sdk.telemetry.Telemetry`. The old top-level function is replaced by `Telemetry.createEvaluationEvent()`.

### Notes
- **Value Parsing**: OpenFeature `Value` definitions natively wrap structures (`Value.Structure`, `Value.List`, etc). Because downstream standard OTEL loggers require primitives, `createEvaluationEvent` employs a robust recursive unwrapping mechanic natively translating custom `dev.openfeature.kotlin.sdk.Value` elements gracefully down to decoupled Kotlin `Map<String, Any?>` interfaces to prevent logging friction.
- This implementation relies on zero external dependencies.

### How to test
`./gradlew :kotlin-sdk:testDebugUnitTest --tests "dev.openfeature.kotlin.sdk.telemetry.*"`


kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/telemetry/Telemetry.kt
```diff
@@ -0,0 +1,107 @@
+package dev.openfeature.kotlin.sdk.telemetry
```
typotter, line 1: We're adding this new telemetry.Telemetry class without deprecating or removing the existing Telemetry class in the parent package. The two classes don't have the same keys (this one matches spec/other SDKs).

Should we update the existing class or remove it? I think it's going to be an acceptable breaking change either way

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/telemetry/Telemetry.kt
```diff
@@ -0,0 +1,107 @@
...
+        }
+
+        // Version
+        val version = evaluationDetails.metadata.getString(TELEMETRY_FLAG_META_VERSION)
+        if (version != null) {
+            attributes[TELEMETRY_VERSION] = version
+        }
+
+        // Error State mapping
+        if (evaluationDetails.reason.equals(Reason.ERROR.name, ignoreCase = true)) {
```
typotter, line 80: `FlagEvaluationDetails.reason` is declared as `String?` (nullable). The `kotlin.text.equals(ignoreCase)` extension has a non-nullable `String` receiver, so calling it on `String?` without `?.` is a compile error — this will not build.

Suggested fix:

```kotlin
if (evaluationDetails.reason?.equals(Reason.ERROR.name, ignoreCase = true) == true)
```

The `?.` safe-call returns `Boolean?`; the `== true` unwraps it so the branch only runs when reason is non-null and matches.

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/telemetry/Telemetry.kt
```diff
@@ -0,0 +1,107 @@
...
+            attributes[TELEMETRY_ERROR_CODE] = evaluationDetails.errorCode?.name ?: ErrorCode.GENERAL.name
```
typotter: `Enum.name` produces uppercase (e.g. `"PROVIDER_FATAL"`), and the tests in this PR assert uppercase — so this is intentional. However, the JS SDK canonical reference lowercases error code values:

```ts
// open-feature/js-sdk packages/shared/src/telemetry/attributes.ts
[ERROR_CODE]: errorCode.toLowerCase(),
```

OTel collectors, dashboards, and queries that join on `error.type` across SDKs will see a mismatch between Kotlin events (`"PROVIDER_FATAL"`) and JS events (`"provider_fatal"`). Is there a deliberate reason to diverge here?

If not, the fix is:
```kotlin
attributes[TELEMETRY_ERROR_CODE] = evaluationDetails.errorCode?.name?.lowercase() ?: ErrorCode.GENERAL.name.lowercase()
```

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/telemetry/EvaluationEvent.kt
```diff
@@ -0,0 +1,9 @@
+package dev.openfeature.kotlin.sdk.telemetry
```
typotter, line 1: The existing `dev.openfeature.kotlin.sdk.EvaluationEvent` (root package) has a `body: Map<String, Any?>` field for the OTel log body. This new `telemetry.EvaluationEvent` drops it with no documentation or migration path.

If any callers populate the log body today, they will silently lose that data when migrated to this class. Either carry the field forward or add a comment documenting the intentional omission and the reasoning.

kotlin-sdk/src/commonTest/kotlin/dev/openfeature/kotlin/sdk/telemetry/TelemetryTest.kt
```diff
@@ -0,0 +1,162 @@
+package dev.openfeature.kotlin.sdk.telemetry
```
typotter, line 1: The implementation falls back to `ErrorCode.GENERAL.name` when `errorCode` is null:

```kotlin
evaluationDetails.errorCode?.name ?: ErrorCode.GENERAL.name
```

Consider adding a test for this case

kotlin-sdk/src/commonTest/kotlin/dev/openfeature/kotlin/sdk/telemetry/TelemetryIntegrationTest.kt
```diff
@@ -0,0 +1,57 @@
...
+
+        val flagResult = client.getBooleanValue("login-button", false)
+        assertFalse(flagResult)
+
+        assertEquals(1, capturedTelemetryEvents.size)
+
+        val metric = capturedTelemetryEvents.first()
+        assertEquals(Telemetry.FLAG_EVALUATION_EVENT_NAME, metric.name)
+        assertEquals("login-button", metric.attributes[Telemetry.TELEMETRY_KEY])
+        // NoOpProvider inherently returns "Passed in default" for variants, skipping telemetry Value propagation explicitly per Appendix D guidelines.
```
typotter: Appendix D footnote [^2] says:

> The `value` field is **required** when a `variant` is not available, and **recommended** when it is.

The spec recommends the inclusion of `value`. It does imply that it is OK without the `value` when `variant` is available, but the instruction here is not explicitly to omit in this case.

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/telemetry/Telemetry.kt
mstepien: That should work, is Null safe. No need to change.
Reason property here is a nullable receiver (String?).
Here is the actual signature `public fun String?.equals(other: String?, ignoreCase: Boolean = false): Boolean`

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/telemetry/Telemetry.kt
```diff
@@ -0,0 +1,110 @@
+package dev.openfeature.kotlin.sdk.telemetry
+
+import dev.openfeature.kotlin.sdk.FlagEvaluationDetails
+import dev.openfeature.kotlin.sdk.HookContext
+import dev.openfeature.kotlin.sdk.Reason
+import dev.openfeature.kotlin.sdk.Value
+import dev.openfeature.kotlin.sdk.exceptions.ErrorCode
+
+/**
+ * The Telemetry object provides constants and utilities for creating OpenTelemetry compliant
```
typotter, line 10: `TELEMETRY_ERROR_MSG` uses `"feature_flag.evaluation.error.message"` but the JS SDK and Appendix D spec both use `"error.message"`.

```kotlin
const val TELEMETRY_ERROR_MSG = "error.message"
```

Ref: JS SDK [`packages/shared/src/telemetry/attributes.ts`](https://github.com/open-feature/js-sdk/blob/8c9669adb08ff5281fd23993fd30701ad665b20a/packages/shared/src/telemetry/attributes.ts) defines `ERROR_MESSAGE = "error.message"`.

kotlin-sdk/src/commonTest/kotlin/dev/openfeature/kotlin/sdk/telemetry/TelemetryTest.kt
```diff
@@ -0,0 +1,216 @@
...
+
+        val event = Telemetry.createEvaluationEvent(hookContext, details)
+
+        val attrs = event.attributes
+        assertEquals("error", attrs[Telemetry.TELEMETRY_REASON])
+        assertEquals("general", attrs[Telemetry.TELEMETRY_ERROR_CODE])
+        assertNull(attrs[Telemetry.TELEMETRY_ERROR_MSG])
+        assertEquals(false, attrs[Telemetry.TELEMETRY_VALUE])
+    }
+}
```
typotter, line 216: Missing test coverage for a few edge cases that existed in the old test suite:

1. **Null provider name** — `ProviderMetadata.name` is `String?`. Old code defaulted to `""`, new code passes `null` through to attributes. Add a test asserting behavior when `providerMetadata.name` is null.

2. **ContextId fallback to targetingKey** — old suite had a test for when metadata `contextId` is absent but `ctx.getTargetingKey()` is set. Current tests only exercise the metadata-present path.

3. **ContextId omitted when both null/empty** — verify `TELEMETRY_CONTEXT_ID` is absent (not null-valued) when both metadata contextId and targetingKey are null/empty.

typotter:
This PR removes `dev.openfeature.kotlin.sdk.EvaluationEvent` and `dev.openfeature.kotlin.sdk.TelemetryKt` (the top-level `createEvaluationEvent` function) from the binary API and replaces them with `dev.openfeature.kotlin.sdk.telemetry.*`. Binary breaking change for any consumer importing the old symbols.

Pre-1.0 SDK so this is fine, but worth noting in the PR description so it shows up in release notes. Something like:

> **Breaking:** `EvaluationEvent` and `createEvaluationEvent()` moved from `dev.openfeature.kotlin.sdk` to `dev.openfeature.kotlin.sdk.telemetry.Telemetry`. The old top-level function is replaced by `Telemetry.createEvaluationEvent()`.

PR #239, opened 2026-04-17
feat!: Remove deprecated ProviderNotReady event

maxnrp:
### Intent
Bring the Kotlin SDK in-line with the OpenFeature specification.

- Canonical events don't include `PROVIDER_NOT_READY`: [5.1.1.](https://openfeature.dev/specification/sections/events/#requirement-511), [5.3.5.](https://openfeature.dev/specification/sections/events/#requirement-535).

- `PROVIDER_ERROR` should use structured details (`message` and `code`): [5.1.4.](https://openfeature.dev/specification/sections/events/#requirement-514), [5.1.5.](https://openfeature.dev/specification/sections/events/#requirement-515).

### Motivation
`PROVIDER_NOT_READY` is not a spec event — NOT_READY is a client/provider status. Keeping `OpenFeatureProviderEvents.ProviderNotReady` invited non-portable provider implementations.
`ProviderError.error` was redundant with `EventDetails` and duplicated how errors are represented compared to the spec’s “provider event details” model.

`fromMessageAndErrorCode` previously dropped the custom message for `PROVIDER_NOT_READY`; using `EventDetails` everywhere makes consistent, message-bearing `ProviderNotReadyError` instances possible.

### Changes
- Removed `OpenFeatureProviderEvents.ProviderNotReady` and updated `MultiProvider` so aggregate `NOT_READY` comes from initial/lifecycle state, not from a synthetic event.

- `ProviderError` now has only `eventDetails`;
- `toOpenFeatureStatusError()` simplified accordingly.
- `OpenFeatureError.fromMessageAndErrorCode: PROVIDER_NOT_READY` now constructs `ProviderNotReadyError(errorMessage)` so the message from `EventDetails` is preserved.

### Tests
- **MultiProviderTests**: precedence scenario for `NOT_READY` over Error/Stale rewritten so one child stays NOT_READY without emitting the removed event; aggregate precedence test updated after dropping `ProviderNotReady` from another scenario.

- **EventDetailsTests**, **ProviderEventingTests**, **AutoHealingProvider**: migrated to `ProviderError(EventDetails(...))` (including `PROVIDER_NOT_READY` with explicit messages).

### Breaking changes
- `ProviderNotReady` event has been removed.
- `ProviderError` no longer accepts an `error` value. Instead, it now only accepts an optional `EventDetails` object (`eventDetails: EventDetails?`).

PR #241, opened 2026-04-22
feat!: Add SMP + Legacy Provider adaptor

maxnrp:
### Intent
Address the race condition in the current SDK implementation, where lifecycle state is split between the provider and the API, causing undefined event ordering in concurrent scenarios.

This is achieved by narrowing event emission to a single source of truth — the provider itself — via the `status: StateFlow<OpenFeatureStatus>` property on `StateManagingProvider`, eliminating the window between `initialize()` returning and the API updating its internal state. State transitions are the authoritative origin of events, which are emitted in response to each lifecycle stage change.

Introduce a small helper for `StateManagingProvider` implementations, the `ProviderStatusTracker`: one place to drive `OpenFeatureStatus` and the provider event stream with the same event → status rules as the rest of the SDK.

### Motivation
[open-feature/spec#365 — Provider lifecycle race condition in multi-threaded SDKs](https://github.com/open-feature/spec/issues/365)
The issue above describes how split ownership of state and events between the API and the provider can produce undefined ordering under concurrency (e.g. API publishing “ready” while the provider emits something else from background work when it actually finishes initializing).

This change aligns the Kotlin SDK with the direction there: the provider is the party that can own a single, authoritative `OpenFeatureStatus` stream. **The API forwards it instead of racing to update the same conceptual state after initialize for implementations**. That does not rewrite the whole spec, but it reduces split ownership for the new path while preserving the old behavior for legacy providers.

Providers that only implement `FeatureProvider` (and not `StateManagingProvider`) are automatically wrapped by `LegacyFeatureProviderAdapter` at registration time. The adapter derives status from the provider's `observe()` events — preserving all existing behavior including status buffering, Ready/Error handling around `initialize()` and `onContextSet()`, and event mirroring — while making the provider fully `StateManagingProvider`-compliant internally. This wrapping is transparent: `getProvider()` always returns the original provider instance.

### Changes
- Introduced `StateManagingProvider` as an extension of `FeatureProvider`, exposing status: `StateFlow<OpenFeatureStatus>` as the single source of truth for lifecycle state.
- Introduced `LegacyFeatureProviderAdapter`, an internal adapter that wraps plain `FeatureProvider` implementations, deriving their status from `observe()` events and preserving legacy lifecycle semantics.
- `OpenFeatureAPI` now:
Delegates `getStatus() `and `statusFlow` directly to the active provider's `status: StateFlow`.
Normalizes all registered providers to `StateManagingProvider` at registration time — either directly if already implementing it, or via `LegacyFeatureProviderAdapter`.
Updates `setProvider` and `setProviderAndWait` to suspend until the provider emits a state other than `NotReady` after `initialize()` completes.
getProvider() unwraps LegacyFeatureProviderAdapter to always return the original registered provider.
- `NoOpProvider` and `MultiProvider` implement `StateManagingProvider`.
- `ProviderStatusTracker` updates derived StateFlow<OpenFeatureStatus> via toOpenFeatureStatus() and broadcasts with tryEmit on a MutableSharedFlow.

### Tests
**StateManagingProviderStatusTests** — Legacy init, observe mirrored into SDK status via LegacyFeatureProviderAdapter, onContextSet, getStatus from provider status, provider replacement/shutdown.

**LegacyFeatureProviderStatusTests** — Legacy init, observe mirrored into SDK `status`, `onContextSet`, `getStatus` from SDK buffer, provider replacement/shutdown.

**StatusTests** — End-to-end transitions, clear/shutdown, concurrent context updates with a state-managing slow provider, provider replacement (including async `setProvider`).

**MultiProviderTests** — Existing multi-provider behavior (strategies, lifecycle, track, error aggregation, init with long-running flows).

**ProviderStatusTrackerTest** — send → status mapping, replay + subsequent sends, not-ready/timeout, two collectors, ProviderConfigurationChanged not changing status, error details preserved under replay, etc.

Together, these unite legacy and state-managing behavior and cover init, status, and context and shutdown, under concurrency where relevant.

### Breaking changes vs legacy compliance
- Intended backward compatible.
- `StateManagingProvider` (standard for custom providers)
- `LegacyFeatureProviderAdapter` (internal — not part of public API)
- API update: new public type: `StateManagingProvider`

## Dependency

This change requires a follow-up merge in: open-feature/kotlin-sdk-contrib

- PR: https://github.com/open-feature/kotlin-sdk-contrib/pull/35


toddbaert:
Related spec PR: https://github.com/open-feature/spec/pull/367

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/events/OpenFeatureProviderEvents.kt
```diff
@@ -45,6 +45,39 @@ sealed class OpenFeatureProviderEvents {
...
+     */
+    data class ProviderReconciling(
+        override val eventDetails: EventDetails? = null
+    ) : OpenFeatureProviderEvents()
+
+    @Deprecated("Use ProviderError instead", ReplaceWith("ProviderError"))
+    data object ProviderNotReady : OpenFeatureProviderEvents() {
+        override val eventDetails = null
+    }
+}
```
dd-oleksii: minor: is this a merge artifact from https://github.com/open-feature/kotlin-sdk/pull/239?

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/events/OpenFeatureProviderEvents.kt
```diff
@@ -45,6 +45,39 @@ sealed class OpenFeatureProviderEvents {
...
+ */
+internal fun OpenFeatureProviderEvents.toOpenFeatureStatus(): OpenFeatureStatus? = when (this) {
+    is OpenFeatureProviderEvents.ProviderReady -> OpenFeatureStatus.Ready
+    is OpenFeatureProviderEvents.ProviderNotReady -> OpenFeatureStatus.NotReady
+    is OpenFeatureProviderEvents.ProviderReconciling -> OpenFeatureStatus.Reconciling
+    is OpenFeatureProviderEvents.ProviderStale -> OpenFeatureStatus.Stale
+    is OpenFeatureProviderEvents.ProviderError -> when (eventDetails?.errorCode) {
+        ErrorCode.PROVIDER_NOT_READY -> OpenFeatureStatus.NotReady
+        else -> toOpenFeatureStatusError()
+    }
```
dd-oleksii: minor: what's the pros of handling `PROVIDER_NOT_READY` here vs in `toOpenFeatureStatusError()`?

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/multiprovider/MultiProvider.kt
```diff
@@ -202,18 +209,14 @@ class MultiProvider(
...
-                return
-            }
-
-            is OpenFeatureProviderEvents.ProviderReady -> OpenFeatureStatus.Ready
-            is OpenFeatureProviderEvents.ProviderStale -> OpenFeatureStatus.Stale
-            is OpenFeatureProviderEvents.ProviderError -> event.toOpenFeatureStatusError()
+        if (event is OpenFeatureProviderEvents.ProviderConfigurationChanged) {
+            // Configuration-only: forward to observers; aggregate readiness is unchanged.
+            eventFlow.emit(event)
+            return
```
dd-oleksii: nitpick: we have somewhat duplicate logic/knowledge here and in `OpenFeatureProviderEvents.toOpenFeatureStatus()` (i.e. that `ProviderConfigurationChanged` does not change status).

Maybe it is slightly cleaner to unify it with the section below:
```kotlin
val newChildStatus = event.toOpenFeatureStatus()
if (newChildStatus == null) {
  eventFlow.emit(event)
  return
}
```

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/multiprovider/MultiProvider.kt
```diff
@@ -264,6 +267,15 @@ class MultiProvider(
...
         }
+        _statusFlow.value = OpenFeatureStatus.NotReady
+        eventFlow.tryEmit(
+            OpenFeatureProviderEvents.ProviderError(
+                OpenFeatureProviderEvents.EventDetails(
+                    message = "MultiProvider shut down; not ready for evaluation",
+                    errorCode = ErrorCode.PROVIDER_NOT_READY
+                )
+            )
+        )
```
dd-oleksii: major: I believe that we should update the status and emit event at the start of `shutdown()` (here and suggest it elsewhere). The logic is that provider loses its ability to evaluate flags _somewhere in the middle_ of shutdown sequence. So it makes sense to flag this proactively/beforehand

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/LegacyFeatureProviderAdapter.kt
```diff
@@ -0,0 +1,101 @@
...
+    /**
+     * [inner] is shut down first while the [observeJob] is still running so a provider can emit
+     * a final [OpenFeatureProviderEvents] on [FeatureProvider.observe] and this adapter can still
+     * apply [toOpenFeatureStatus] to [_status]. The observe job and [scope] (including
+     * [SupervisorJob]) are cancelled in [finally] so cleanup and resource release always run.
+     */
+    override fun shutdown() {
+        try {
+            inner.shutdown()
+            _status.value = OpenFeatureStatus.NotReady
```
dd-oleksii, line 75: major: similarly here — we need to update the status and emit the event before calling the `shutdown()`

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/LegacyFeatureProviderAdapter.kt
```diff
@@ -0,0 +1,101 @@
...
+        } catch (e: Throwable) {
+            handleError(e)
```
dd-oleksii, line 87: opinion: I'd argue that we should ignore errors during the shutdown and providers should generally not throw during shutdown. Mainly because we can't do anything about it: we are shutting down and it's an opportunity for the provider to cleanup — if they can't cleanup, too bad but we're moving on 🤷

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/LegacyFeatureProviderAdapter.kt
```diff
@@ -0,0 +1,101 @@
...
+            _status.value = OpenFeatureStatus.Ready
+        } catch (e: Throwable) {
+            handleError(e)
+        }
+    }
+
+    private fun handleError(e: Throwable) {
+        when (e) {
+            is CancellationException -> { /* Cancelled by design - not an error */ }
+            is OpenFeatureError -> _status.value = OpenFeatureStatus.Error(e)
```
dd-oleksii, line 111: major: how do we handle fatal or not ready errors?

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/LegacyFeatureProviderAdapter.kt
```diff
@@ -0,0 +1,101 @@
...
+) : StateManagingProvider,
+    FeatureProvider by inner {
+
+    private val _status = MutableStateFlow<OpenFeatureStatus>(OpenFeatureStatus.NotReady)
+    override val status: StateFlow<OpenFeatureStatus> = _status.asStateFlow()
+
+    private val scope = CoroutineScope(SupervisorJob() + eventDispatcher)
+    private var observeJob: Job? = null
+
+    override fun observe(): Flow<OpenFeatureProviderEvents> = inner.observe()
```
dd-oleksii: major: we can't directly expose inner events because this is missing all the events that were previously emitted by SDK. We're also missing these event emissions in `initialize`/`shutdown`/`onContextSet`.

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/OpenFeatureAPI.kt
```diff
@@ -81,59 +104,57 @@ object OpenFeatureAPI {
...
         }
 
-        // Emit NotReady status after swapping provider
-        _statusFlow.emit(OpenFeatureStatus.NotReady)
-
         // Shutdown the previous provider outside the mutex
-        tryWithStatusEmitErrorHandling {
-            oldProvider.shutdown()
-        }
+        oldProvider.shutdown()
```
dd-oleksii, line 128: major: do we need try-catch here?

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/OpenFeatureAPI.kt
```diff
@@ -81,59 +104,57 @@ object OpenFeatureAPI {
...
 
-        // Initialize the new provider
-        tryWithStatusEmitErrorHandling {
-            listenToProviderEvents(provider, dispatcher)
-            getProvider().initialize(context)
-            _statusFlow.emit(OpenFeatureStatus.Ready)
+        normalizedProvider.initialize(context)
+        normalizedProvider.status.first {
+            it !is OpenFeatureStatus.NotReady && it !is OpenFeatureStatus.Reconciling
         }
```
dd-oleksii, line 133: Q: what is this doing?

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/OpenFeatureAPI.kt
```diff
@@ -81,59 +104,57 @@ object OpenFeatureAPI {
...
     }
 
     /**
-     * Get the current [FeatureProvider] for the SDK.
+     * Get the current [FeatureProvider] registered by the application.
+     *
+     * When a plain [FeatureProvider] was set, this returns that instance (not an internal wrapper).
```
dd-oleksii, line 139: minor: this shouldn't be in the public-facing doc because users don't know about internal wrappers
```suggestion
```

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/OpenFeatureAPI.kt
```diff
@@ -177,30 +198,9 @@ object OpenFeatureAPI {
...
-        } catch (e: Throwable) {
-            _statusFlow.emit(
-                OpenFeatureStatus.Error(
-                    OpenFeatureError.GeneralError(
-                        e.message ?: "Unknown error"
-                    )
-                )
-            )
+        if (oldContext != evaluationContext) {
+            provider.onContextSet(oldContext, evaluationContext)
```
dd-oleksii, line 203: major: try-catch?

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/multiprovider/MultiProvider.kt
```diff
@@ -55,7 +58,7 @@ typealias FlagEval<T> =
 class MultiProvider(
```
dd-oleksii, line 54: minor: does it make sense to use `ProviderStatusTracker` in `MultiProvider`?

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/StateManagingProvider.kt
```diff
@@ -0,0 +1,69 @@
...
+import kotlin.coroutines.cancellation.CancellationException
+
+/**
+ * Provider that owns lifecycle [OpenFeatureStatus] in [status] and reports changes on [observe].
+ *
+ * Implementations must update [status] first, then emit on [observe], so readers of either see a
+ * consistent order. Plain [FeatureProvider] implementations are wrapped internally; their status is
+ * derived from [observe] via an SDK adapter.
+ */
+interface StateManagingProvider : FeatureProvider {
```
dd-oleksii, line 16: minor: shall we put deprecation on `FeatureProvider`?

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/StateManagingProvider.kt
```diff
@@ -0,0 +1,69 @@
...
+     * @param initialContext optional initial [EvaluationContext]
+     */
+    @Throws(OpenFeatureError::class, CancellationException::class)
+    override suspend fun initialize(initialContext: EvaluationContext?)
+
+    /**
+     * Called when the client lifecycle ends; release resources and threads.
+     *
+     * Before returning: set [status] to [OpenFeatureStatus.NotReady], then emit
+     * [OpenFeatureProviderEvents.ProviderError] on [observe].
```
dd-oleksii: nitpick: clarify that we expect `ProviderError` with `NotReady` error code.

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/StateManagingProvider.kt
```diff
@@ -0,0 +1,69 @@
...
+    @Throws(OpenFeatureError::class, CancellationException::class)
+    override suspend fun onContextSet(oldContext: EvaluationContext?, newContext: EvaluationContext)
+
+    /**
+     * Provider lifecycle events. Replace [FeatureProvider.observe]'s empty default.
+     *
+     * Implementations update [status] before each matching emission. For [initialize], set [status]
+     * first, then emit [OpenFeatureProviderEvents.ProviderReady] or
+     * [OpenFeatureProviderEvents.ProviderError], or the SDK remains [OpenFeatureStatus.NotReady]. Other
+     * events surface through [OpenFeatureAPI.observe].
```
dd-oleksii: nitpick: not sure why `initialize` is cherry-picked here and not user what `Other events surface through [OpenFeatureAPI.observe]` means

kotlin-sdk/src/commonTest/kotlin/dev/openfeature/kotlin/sdk/helpers/LegacyMinimalProvider.kt
```diff
@@ -0,0 +1,76 @@
...
+    override val metadata: ProviderMetadata = object : ProviderMetadata {
+        override val name: String? = "legacy-minimal"
+    }
+) : FeatureProvider {
+    val shutdownCalls = atomic(0)
+
+    private val events = MutableSharedFlow<OpenFeatureProviderEvents>(replay = 1, extraBufferCapacity = 5)
+
+    override suspend fun initialize(initialContext: EvaluationContext?) {
+        events.emit(OpenFeatureProviderEvents.ProviderReady())
```
dd-oleksii: major: legacy providers should _not_ do that because SDK was emitting it for them

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/events/OpenFeatureProviderEvents.kt
maxnrp: that's garbage, thank you, removed.

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/events/OpenFeatureProviderEvents.kt
maxnrp: well, I don’t really understand `OpenFeatureStatus.NotReady` being a state reached through an error type event. Not being ready seems like something expected and part of the natural flow, not particularly an error.

so, maybe the wish to handle them here is me seeing it like that. But I think handling it inside `toOpenFeatureStatusError()` would be the most correct approach considering the current spec. Updated.

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/multiprovider/MultiProvider.kt
maxnrp: I was trying to that in separate PR to facilitate reading, but seems simpler than I thought, so, done, ty.

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/events/OpenFeatureProviderEvents.kt
dd-oleksii: I share your views :) https://github.com/open-feature/spec/issues/366

I believe the reason for using error is because errors existed before events, and `PROVIDER_NOT_READY` is a way to signal that evaluation failed because provider isn't ready yet. So given the existence of that error, it makes sense to set provider status to `NotReady` if we receive it... but I agree that this is a rather convoluted way to signal not-readiness, and a standalone "not ready" event would be preferred. It was less of an issue before because providers never had to signal their not-readiness, so `event(error=not ready)` was never used but not providers have to emit it on shutdown

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/events/OpenFeatureProviderEvents.kt
```diff
@@ -45,12 +45,34 @@ sealed class OpenFeatureProviderEvents {
...
     ) : OpenFeatureProviderEvents()
+
+    /**
+     * The provider is in the process of reconciling within a context change.
+     */
+    data class ProviderReconciling(
+        override val eventDetails: EventDetails? = null
+    ) : OpenFeatureProviderEvents()
+}
+
```
nicklasl, line 56: Can we add a `ProviderContextChanged` event type here? The [updated spec (5.1.1)](https://github.com/open-feature/spec/blob/main/specification/sections/05-events.md#requirement-511) made PROVIDER_CONTEXT_CHANGED a MUST-level event that maps to READY status — I guess we'll want to add it to `toOpenFeatureStatus()` too so the mapping table is complete before this lands.

PR #243, opened 2026-04-24
feat!: add isolated api instances

marcozabel:
## This PR
This PR introduces support for creating isolated OpenFeature API instances, each with their own providers, hooks, context, and event handling - enabling multi-tenant or side-by-side usage without shared global state.

### Related Issues
https://github.com/open-feature/kotlin-sdk/issues/229


kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/OpenFeatureAPI.kt
```diff
@@ -1,292 +1,20 @@
...
-            .apply {
-                tryEmit(OpenFeatureStatus.NotReady)
-            }
-
-    val statusFlow: Flow<OpenFeatureStatus> get() = _statusFlow.distinctUntilChanged()
-
-    var hooks: List<Hook<*>> = listOf()
-        private set
-
+object OpenFeatureAPI : OpenFeatureInstance() {
```
bencehornak: I would prefer `val OpenFeatureAPI = OpenFeatureInstance()`, because this is not an actual inheritance (no behavior is overridden).

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/OpenFeatureAPI.kt
```diff
@@ -1,292 +1,20 @@
...
-
-            is OpenFeatureProviderEvents.ProviderError -> {
-                _statusFlow.emit(providerEvent.toOpenFeatureStatusError())
-            }
-
-            else -> { // All other states should not be emitted from here
-            }
-        }
-    }
+    fun createInstance(): OpenFeatureInstance = OpenFeatureInstance()
```
bencehornak: This can become a top-level method in another file outside of any class/object declarations.

Related: https://github.com/open-feature/spec/blob/main/specification/sections/01-flag-evaluation.md#requirement-183

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/OpenFeatureInstance.kt
```diff
@@ -0,0 +1,361 @@
...
+ * Core implementation of the OpenFeature API.
+ *
+ * Each instance maintains its own independent state: provider, evaluation context, hooks, status,
+ * and events. The global singleton [OpenFeatureAPI] extends this class. To create isolated,
+ * independent instances use [OpenFeatureAPI.createInstance].
+ *
+ * @see OpenFeatureAPI
+ */
+@Suppress("TooManyFunctions")
+open class OpenFeatureInstance {
```
bencehornak:
```suggestion
class OpenFeatureInstance internal constructor {
```

bencehornak (commented):
I usually find it a good practice to make only interfaces public and the implementing classes internal.

```kotlin
val OpenFeatureAPI: OpenFeatureAPIInterface = OpenFeatureImpl()
fun createIsolatedOpenFeatureAPI(): OpenFeatureAPIInterface = OpenFeatureImpl()

interface OpenFeatureAPIInterface {...}
private /* or internal */ class OpenFeatureImpl {...}
```

This is what the official Kotlin libs often do too:
```kotlin
public interface MutableStateFlow<T> : StateFlow<T>, MutableSharedFlow<T> {...}
public fun <T> MutableStateFlow(value: T): MutableStateFlow<T> = StateFlowImpl(value ?: NULL)
private class StateFlowImpl<T> {...}
```

(From https://github.com/Kotlin/kotlinx.coroutines/blob/1.10.2/kotlinx-coroutines-core/common/src/flow/StateFlow.kt)

This allows separating what's public and what's internal. Additionally, it's easier to mock the interfaces in unit tests than `object`s.

bencehornak:
If it was a green field project, I'd suggest the names:

```kotlin
val OpenFeature: OpenFeatureAPI = OpenFeatureImpl() // BREAKING CHANGE
fun createIsolatedOpenFeatureAPI(): OpenFeatureAPI = OpenFeatureImpl()

interface OpenFeatureAPI {...}
private /* or internal */ class OpenFeatureImpl {...}
```

instead of the names I used in my previous comment:

```kotlin
val OpenFeatureAPI: OpenFeatureAPIInterface = OpenFeatureImpl()
fun createIsolatedOpenFeatureAPI(): OpenFeatureAPIInterface = OpenFeatureImpl()

interface OpenFeatureAPIInterface {...}
private /* or internal */ class OpenFeatureImpl {...}
```

But it's a breaking change, so I'm not sure if it's worth annoying all current lib users.

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/OpenFeatureInstance.kt
```diff
@@ -0,0 +1,356 @@
...
+        }
+
+        // Emit NotReady status after swapping provider
+        _statusFlow.emit(OpenFeatureStatus.NotReady)
+
+        // Shutdown the previous provider outside the mutex
+        tryWithStatusEmitErrorHandling {
+            untrackProviderBinding(oldProvider)
+            oldProvider.shutdown()
+        }
```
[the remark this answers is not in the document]
marcozabel: Already addressed — the `if (oldProvider !== provider)` guard is in place on line 127, along with a test (`testReSettingSameProviderDoesNotShutItDown`).

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/OpenFeatureAPIInstance.kt
```diff
@@ -0,0 +1,356 @@
...
+     * Clear the current [FeatureProvider] and reset to a no-op provider.
+     */
+    suspend fun clearProvider() {
+        val oldProvider = provider
+        provider = noOpProvider
+        providersFlow.value = noOpProvider
+        untrackProviderBinding(oldProvider)
+        oldProvider.shutdown()
+        _statusFlow.emit(OpenFeatureStatus.NotReady)
+    }
```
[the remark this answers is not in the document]
marcozabel: Already addressed — `clearProvider()` uses `providerMutex.withLock` (line 153) to atomically swap the provider before untrack/shutdown.

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/OpenFeatureInstance.kt
marcozabel: Agreed, making the constructor `internal` prevents arbitrary external subclassing while keeping the type public for return values. Will update.

marcozabel:
Thanks for the review @bencehornak!

interface + internal impl pattern: I agree that's the ideal Kotlin pattern for new APIs. For this PR though, `OpenFeatureAPI` is an existing public `object` with consumers depending on `OpenFeatureAPI.INSTANCE` at the JVM bytecode level. Introducing an interface and hiding the impl would be a larger breaking change that deserves its own discussion/PR. Happy to explore that as a follow-up.

I updated moving `createInstance()` to a top-level function and making the `OpenFeatureInstance` constructor `internal`. 👍

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/OpenFeatureAPI.kt
marcozabel: You're right, no behavior is overridden. We keep object for binary compatibility because it generates a JVM class with a static final INSTANCE field that existing compiled consumers reference. A top-level val removes that class entirely causing NoClassDefFoundError at runtime.

bencehornak:
@marcozabel I think breaking changes are still allowed, esp. in the jvm compatibility layer, since we are pre 1.0.

However, you are right, we should try to avoid them, if possible. One more thought on my previous line of thought: with the following trick both the Kotlin (`OpenFeatureAPI.xxx`) and the JVM signatures (`OpenFeatureAPI.INSTANCE.xxx`) can be the same as before:

```kotlin
@file:JvmName("OpenFeatureAPI")

@JvmName("INSTANCE")
val OpenFeatureAPI = OpenFeatureInstance()
```

IMO this pattern describes the specs better than the current proposal, because the convenience OpenFeatureAPI is an instance of `OpenFeatureInstance`, just like any other instance returned by the factory method.

And this pattern would allow the smooth introduction of the separation between the interface and implementation classes in the same PR without noticeable impact from the outside.

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/OpenFeatureInstance.kt
```diff
@@ -0,0 +1,362 @@
...
+ *
+ * Each instance maintains its own independent state: provider, evaluation context, hooks, status,
+ * and events. The global singleton [OpenFeatureAPI] extends this class. To create isolated,
+ * independent instances use [createOpenFeatureInstance].
+ *
+ * @see OpenFeatureAPI
+ * @see createOpenFeatureInstance
+ */
+@Suppress("TooManyFunctions")
+open class OpenFeatureInstance internal constructor() {
```
bencehornak: Let's use a more specific class name, to indicate that it's an instance of the API (and not a Provider or a Client). So how about OpenFeatureAPIInstance?

Have you checked what class names the SDKs of other languages have introduced?

kotlin-sdk/src/commonTest/kotlin/dev/openfeature/kotlin/sdk/IsolatedAPIInstanceTests.kt
```diff
@@ -0,0 +1,367 @@
...
+
+        OpenFeatureAPI.addHooks(listOf(hook1))
+        instance.addHooks(listOf(hook2))
+
+        assertEquals(1, OpenFeatureAPI.hooks.size)
+        assertEquals(1, instance.hooks.size)
+        assertTrue(OpenFeatureAPI.hooks.contains(hook1))
+        assertFalse(OpenFeatureAPI.hooks.contains(hook2))
+        assertTrue(instance.hooks.contains(hook2))
+        assertFalse(instance.hooks.contains(hook1))
```
bencehornak:
```suggestion
        assertEquals(listOf(hook1)), OpenFeatureAPI.hooks)
        assertEquals(listOf(hook2), instance.hooks)
```

kotlin-sdk/src/commonTest/kotlin/dev/openfeature/kotlin/sdk/IsolatedAPIInstanceTests.kt
```diff
@@ -0,0 +1,367 @@
...
+    private val instances = mutableListOf<OpenFeatureInstance>()
+
+    private fun createInstance(): OpenFeatureInstance {
+        val instance = createOpenFeatureInstance()
+        instances.add(instance)
+        return instance
+    }
+
+    @AfterTest
+    fun tearDown() = runTest {
```
bencehornak, line 31: I'd reset the boundProviders field here to make sure that the test cases have no side-effects. Can you add a test-only internal clearBoundProviders method to the companion object, which eould be called from here?

bencehornak (approved):
Looks good otherwise, see my suggestions above.

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/OpenFeatureInstance.kt
marcozabel: Yes, I implemented this across some SDKs. For most of them I added a new Factory to generate the instance and imho that made it implicit 😄

kotlin-sdk/src/commonTest/kotlin/dev/openfeature/kotlin/sdk/IsolatedAPIInstanceTests.kt
marcozabel: I did not apply the suggestion, because then you wouldn't be allowed to approve anymore. But I implemented it 👍

kotlin-sdk/api/android/kotlin-sdk.api
```diff
@@ -272,12 +272,17 @@ public final class dev/openfeature/kotlin/sdk/NoOpProvider$NoOpProviderMetadata
 }
 
 public final class dev/openfeature/kotlin/sdk/OpenFeatureAPI {
-	public static final field INSTANCE Ldev/openfeature/kotlin/sdk/OpenFeatureAPI;
+	public static final fun INSTANCE ()Ldev/openfeature/kotlin/sdk/OpenFeatureAPIInstance;
```
bencehornak: Ah sorry, forgot to say that we need the `@JvmField` annotation too, otherwise it's compiled as a getter method.

bencehornak (approved):
One more thing: would you mind adding a Kotlin and a Java snippet to the README about the two usage patterns? (Singleton+factory)

kotlin-sdk/api/android/kotlin-sdk.api
bencehornak: Optional idea: to prevent accidental changes to the java api, we could add some superficial test cases written in java under `kotlin-sdk/src/jvmTest/java` (or sth like this)

bencehornak (changes requested):
https://github.com/open-feature/kotlin-sdk/pull/243#discussion_r3160419943

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/OpenFeatureAPI.kt
```diff
@@ -12,8 +13,10 @@ import kotlin.jvm.JvmName
  *
  * This is an instance of [OpenFeatureAPIInstance], just like any instance returned by
  * the [createOpenFeatureAPIInstance] factory method.
+ *
+ * @apiNote Isolated API instances (per spec section 1.8) are experimental and subject to change.
  */
-@get:JvmName("INSTANCE")
+@JvmField
```
bencehornak: TIA, these two are apparently mutually exclusive. This way the the singleton will need to be accessed like `OpenFeatureAPI.OpenFeatureAPI` from Java code (as opposed to `OpenFeatureAPI.INSTANCE`) in versions after this:

```diff
public final class dev/openfeature/kotlin/sdk/OpenFeatureAPI {
-	 public static final fun INSTANCE ()Ldev/openfeature/kotlin/sdk/OpenFeatureAPIInstance;
+	 public static final field OpenFeatureAPI Ldev/openfeature/kotlin/sdk/OpenFeatureAPIInstance;
  ...
}
```

I'm fine with this, and I wouldn't worry too much about changing the JVM binary interface. We are in the 0.x range, where every version [is allowed](https://semver.org/#spec-item-4) to contain breaking changes according to the semver specs.

bencehornak (approved):

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/OpenFeatureAPI.kt
```diff
@@ -1,292 +1,28 @@
...
+val OpenFeatureAPI: OpenFeatureAPIInstance = OpenFeatureAPIInstance()
+
+/**
+ * Create a new, independent [OpenFeatureAPIInstance] with its own provider, context, hooks,
+ * and events — completely isolated from the global [OpenFeatureAPI] singleton and other instances.
+ *
+ * @return a new [OpenFeatureAPIInstance]
+ */
+fun createOpenFeatureAPIInstance(): OpenFeatureAPIInstance = OpenFeatureAPIInstance()
```
bencehornak: nit: `@apiNote` is javadoc, afaik it doesn't work with kdoc. And the note rather belongs to the factory method, doesn't it?

```suggestion
/**
 * Global singleton entry point for the OpenFeature SDK.
 *
 * Use this directly for typical single-provider usage. For isolated, independent instances
 * (e.g., for DI frameworks or testing), use [createOpenFeatureAPIInstance].
 *
 * This is an instance of [OpenFeatureAPIInstance], just like any instance returned by
 * the [createOpenFeatureAPIInstance] factory method.
 */
@JvmField
val OpenFeatureAPI: OpenFeatureAPIInstance = OpenFeatureAPIInstance()

/**
 * Create a new, independent [OpenFeatureAPIInstance] with its own provider, context, hooks,
 * and events — completely isolated from the global [OpenFeatureAPI] singleton and other instances.
 *
 * **Note:** isolated API instances (per spec section 1.8) are experimental and subject to change.
 *
 * @return a new [OpenFeatureAPIInstance]
 */
fun createOpenFeatureAPIInstance(): OpenFeatureAPIInstance = OpenFeatureAPIInstance()
```

toddbaert:
Pushed a commit moving `createOpenFeatureAPIInstance` into a `dev.openfeature.kotlin.sdk.isolated` sub-package and relocating the `@apiNote` experimental marker onto `OpenFeatureAPIInstance` and the factory itself per spec 1.8.3; @marcozabel @bencehornak mind taking another look?

toddbaert (approved):

toddbaert:
> Pushed a commit moving `createOpenFeatureAPIInstance` into a `dev.openfeature.kotlin.sdk.isolated` sub-package and relocating the `@apiNote` experimental marker onto `OpenFeatureAPIInstance` and the factory itself per spec 1.8.3; @marcozabel @bencehornak mind taking another look?

I also rebased on main and fixed conflicts.

toddbaert:
@dd-oleksii @nicklasl any opposition to this? I will merge Friday unless I hear an objection from you.

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/OpenFeatureAPIInstance.kt
```diff
@@ -0,0 +1,377 @@
...
+
+    @OptIn(ExperimentalCoroutinesApi::class)
+    private suspend fun setProviderInternal(
+        provider: FeatureProvider,
+        dispatcher: CoroutineDispatcher,
+        initialContext: EvaluationContext? = null
+    ) {
+        try {
+            trackProviderBinding(provider)
```
nicklasl, line 108: nit: Just wondering: if `setProvider`(not andWait) is called twice quickly and the first coroutine gets cancelled between this place and the `untrackProviderBinding` at line 134, would that provider entry stick around in boundProviders? Probably a pretty narrow window in practice, but I guess a try-catch around the "atomical swap" to catch (and re-throw) the `CancellationException` could help?

nicklasl (approved):

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/OpenFeatureAPIInstance.kt
toddbaert: Yep, added `try/catch` around the post-trackProviderBinding section that reverts if cancellation hits before the swap commits. I tried to write a test for this but couldn't find a way to reasonably trigger the narrow window you describe.

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/OpenFeatureAPI.kt
toddbaert: Thanks, made that change to just prose.

toddbaert:
Pushed a new experimental marker which observes [these conventions](https://kotlinlang.org/docs/opt-in-requirements.html#opt-in-to-api). Note: this PR is technically breaking; `OpenFeatureAPI.INSTANCE` is gone (now `OpenFeatureAPI.OpenFeatureAPI` via `@JvmField` val), so Java consumers need updating. Should probably be feat!: on the squash.

Again, will merge in a day or two.

nicklasl (approved):

Issue #245, opened 2026-04-28
Centralize event distribution in OpenFeatureAPI.observe() to avoid redundant provider subscriptions


typotter:
## Problem

Every call to `.collect` on a flow returned by `OpenFeatureAPI.observe()` (or `Client.observe()`) creates an independent `flatMapLatest` subscription into the underlying `provider.observe()`:

```kotlin
// Current implementation
providersFlow.flatMapLatest { it.observe() }.filterIsInstance()
```

With N callers collecting concurrently, there are N independent subscriptions to the provider's event flow. For providers where `observe()` is expensive (e.g. opens a connection, registers a listener), this causes unnecessary overhead.

## Proposed fix

Distribute events from a single upstream subscription using a `SharedFlow`:

```kotlin
private val _providerEvents: SharedFlow<OpenFeatureProviderEvents> = providersFlow
    .flatMapLatest { it.observe() }
    .shareIn(apiScope, SharingStarted.Eagerly, replay = 0)

inline fun <reified T : OpenFeatureProviderEvents> observe(): Flow<T> =
    _providerEvents.filterIsInstance()
```

This keeps one upstream subscription regardless of how many callers collect.

## Context

Surfaced during review of #234 (`Client.observe()`). The fix should live in `OpenFeatureAPI.observe()` so both the API-level and client-level observe paths benefit. Fixing only at the `Client` level would create asymmetry.

PR #246, opened 2026-04-30
feat: Centralize event distribution

maxnrp:
### Intent
Fix redundant `FeatureProvider.observe()` subscriptions when multiple callers collect `OpenFeatureAPI.observe()` / `Client.observe()` ([#245](https://github.com/open-feature/kotlin-sdk/issues/245)).

### Motivation
Each collector used to drive its own `flatMapLatest` into `provider.observe()` — which can be an overhead for providers where `observe()` is costly.

### Changes

- `shareIn` on `providersFlow.flatMapLatest { it.observe() }.shareIn(providerEventsScope, SharingStarted.Eagerly, replay = 0)`.

### Tests
`flushDispatchersDefault(durationMs)`: `runTest` doesn’t pump `Dispatchers.Default`, where `shareIn` runs.
Existing event-order / client–API parity / filter tests kept; they still define expected behavior.

### Breaking changes
None intended. Concurrent collectors now share one hot stream (replay = 0);


kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/OpenFeatureAPI.kt
```diff
@@ -44,6 +58,35 @@ object OpenFeatureAPI {
...
+                _statusFlow.emit(providerEvent.toOpenFeatureStatusError())
+            }
+
+            else -> { // All other states should not be emitted from here
+            }
+        }
+    }
+
+    init {
+        providerEventsScope.launch {
```
bencehornak: cancel this job in `shutdown()`

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/OpenFeatureAPI.kt
```diff
@@ -72,26 +115,19 @@ object OpenFeatureAPI {
      *
      * @param provider the [FeatureProvider] to set
      * @param initialContext the initial [EvaluationContext] to use for the provider initialization. Defaults to an null context if not set.
+     * @param dispatcher retained for API compatibility (no longer used for a separate provider observe job).
```
bencehornak: I'd rather remove this, to not carry tech debt along and to move towards a clean 1.0 API.

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/OpenFeatureAPI.kt
```diff
@@ -14,25 +14,39 @@ import kotlinx.coroutines.flow.Flow
...
     private val providerMutex = Mutex()
     private val NOOP_PROVIDER = NoOpProvider()
     private var provider: FeatureProvider = NOOP_PROVIDER
     private var context: EvaluationContext? = null
     val providersFlow: MutableStateFlow<FeatureProvider> = MutableStateFlow(NOOP_PROVIDER)
 
+    /**
+     * [Dispatchers.Unconfined] keeps the shared [observe] collector on the emitting thread when possible.
+     */
+    private val providerEventsScope = CoroutineScope(SupervisorJob() + Dispatchers.Unconfined)
```
bencehornak: We should not hardcode any dispatchers to make the code testable. In #243 this `object` will become a `class`, so we could take the opportunity to set the `Dispatcher` as a constructor argument.

bencehornak (changes requested):
First batch of thoughts :)

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/OpenFeatureAPI.kt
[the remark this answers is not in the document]
maxnrp: I've tried with `Default` before (can check gemini's insights on this marked as resolved above), but tests looked messy, and I thought that `Unconfined` wasn't a bad option, may seen unstable but the jumping from thread to thread is kinda appealing, isn't?

Though as you suggested in a comment, now that API is becoming a class we could just pass `Dispatcher` as an argument eventually. I've updated the tests, just to keep the PR merge-ready, if #243 goes first, we may retouch it again.

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/OpenFeatureAPI.kt
```diff
@@ -14,25 +14,39 @@ import kotlinx.coroutines.flow.Flow
...
+
+    @PublishedApi
+    @OptIn(ExperimentalCoroutinesApi::class)
+    internal val sharedProviderEvents =
+        providersFlow
+            .flatMapLatest { it.observe() }
+            .shareIn(providerEventsScope, SharingStarted.Eagerly, replay = 0)
```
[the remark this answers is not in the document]
maxnrp: PR description fixed. We always collect sharedProviderEvents in init for status.

PR #249, opened 2026-07-29
feat: add iosX64 target

arminas-vil:
Add the iosX64 Kotlin/Native target so the SDK can be consumed from the iOS simulator on Intel-based Macs, alongside the existing iosArm64 and iosSimulatorArm64 targets.

Also remove the duplicate iosArm64()/iosSimulatorArm64() target declarations that were listed twice in the kotlin block.


AleksandrasMaliuginas-vinted (approved):

arminas-vil:
@coderabbitai review

PR #251, opened 2026-08-06
fix: reconcile evaluation context on every set

btthomas:
Hi friends! Long-time user, first-time contributor here.

I was helping a customer debug stale flag values on Android when I noticed that repeated `setEvaluationContext` calls with the same values were not reaching the provider's `onContextSet` callback. I tried to keep the fix focused and added a few tests to cover it. Let me know if there is anything else I need to do to get this ready to merge. Thanks!

### What changed

The Kotlin SDK skips `FeatureProvider.onContextSet` when the new evaluation context is equal to the current context.

This means that calling `setEvaluationContext` with the same user and attributes does not give the provider a chance to refresh its state. For providers that fetch flag configuration during `onContextSet`, this can leave flag values stale.

This behavior was reported with Kotlin SDK 0.6.2, but the same equality check is still present on `main`.

### SDK behavior before this change

| SDK | Same instance | New object, same values | Changed values |
|---|---:|---:|---:|
| Kotlin 0.6.2 | No callback | No callback | Callback |
| Swift 0.5.0 | Callback | Callback | Callback |
| Web 1.10.0 | Callback | Callback | Callback |

The OpenFeature specification says:

> When the global `evaluation context` is set, the `on context changed` function **MUST** run.

See [Conditional Requirement 3.2.4.1](https://github.com/open-feature/spec/blob/d5b0a734d8cb9b42bf89be2a97c627f58208e811/specification/sections/03-evaluation-context.md#L125-L129).

The requirement is based on calling the setter. It does not make the callback conditional on context equality.

### Implementation

This PR:

- Calls `onContextSet` for every evaluation-context setter call.
- Keeps the provider in `RECONCILING` until all active context updates finish.
- Prevents an old provider callback from changing status after the provider has been replaced or cleared.
- Keeps the SDK `NOT_READY` when context is set without a configured provider.
- Documents that providers may receive equal contexts or the same context instance.

There are no public API signature changes.

### Tests

Added coverage for:

- Setting the same context instance more than once.
- Setting different context objects with equal values.
- The `READY` -> `RECONCILING` -> `READY` lifecycle.
- Overlapping context updates.
- Setting context after the provider has been cleared.

### DIFF
--- a/kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/OpenFeatureAPIInstance.kt
+++ b/kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/OpenFeatureAPIInstance.kt
@@ -209,12 +209,10 @@ open class OpenFeatureAPIInstance internal constructor() {
     private suspend fun setEvaluationContextInternal(evaluationContext: EvaluationContext) {
         val oldContext = context
         context = evaluationContext
-        if (oldContext != evaluationContext) {
-            _statusFlow.emit(OpenFeatureStatus.Reconciling)
-            tryWithStatusEmitErrorHandling {
-                getProvider().onContextSet(oldContext, evaluationContext)
-                _statusFlow.emit(OpenFeatureStatus.Ready)
-            }
+        _statusFlow.emit(OpenFeatureStatus.Reconciling)
+        tryWithStatusEmitErrorHandling {
+            getProvider().onContextSet(oldContext, evaluationContext)
+            _statusFlow.emit(OpenFeatureStatus.Ready)
         }
     }
 
--- a/kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/OpenFeatureAPIInstance.kt
+++ b/kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/OpenFeatureAPIInstance.kt
@@ -54,6 +54,10 @@ open class OpenFeatureAPIInstance internal constructor() {
     private var context: EvaluationContext? = null
     private var contextReconciliationGeneration: Long? = null
     private var activeContextReconciliations: Int = 0
+    private var contextReconciliationInitialStatus: OpenFeatureStatus? = null
+    private var contextReconciliationTerminalStatus: OpenFeatureStatus? = null
+    private var providerStatusGeneration: Long = 0
+    private var contextReconciliationTerminalProviderStatusGeneration: Long? = null
     val providersFlow: MutableStateFlow<FeatureProvider> = MutableStateFlow(noOpProvider)
 
     private val _statusFlow: MutableSharedFlow<OpenFeatureStatus> =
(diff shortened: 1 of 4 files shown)


kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/OpenFeatureAPIInstance.kt
```diff
@@ -207,13 +219,70 @@ open class OpenFeatureAPIInstance internal constructor() {
...
+        } finally {
+            if (reconciliationRegistered) {
+                withContext(NonCancellable) {
+                    completeContextReconciliation(
+                        reconciliation.provider,
+                        reconciliation.providerGeneration,
+                        terminalStatus
+                    )
+                }
+            }
```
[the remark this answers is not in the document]
btthomas: Great feedback. I addressed it in this commit and added a couple tests: https://github.com/open-feature/kotlin-sdk/pull/251/commits/53e76ba390a6084cc88921d419fea4ffaec247b3

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/OpenFeatureAPIInstance.kt
```diff
@@ -36,14 +36,28 @@ import kotlinx.coroutines.withContext
...
     private val noOpProvider = NoOpProvider()
     private var provider: FeatureProvider = noOpProvider
+    private var providerGeneration: Long = 0
     private var context: EvaluationContext? = null
+    private var contextReconciliationGeneration: Long? = null
+    private var activeContextReconciliations: Int = 0
+    private var contextReconciliationInitialStatus: OpenFeatureStatus? = null
+    private var contextReconciliationTerminalStatus: OpenFeatureStatus? = null
+    private var providerStatusGeneration: Long = 0
+    private var contextReconciliationTerminalProviderStatusGeneration: Long? = null
```
toddbaert, line 63: 2 thoughts on this:

All these fields seem to exist in service of supporting interleaving `setEvaluationContexts` on the same instance. If I'm understanding correctly, the simpler solution might be to hold a single mutex across the entire operation, including the `onContextSet` callback. This would serialize context-updates, but I wonder if that's acceptable? In the client paradigm it's not an operation we'd expect in a tight loop; it's driven by a significant user action(s) that impacts flag evaluations (login, logout, change subscription level, etc.).

If we decide we do need to support this re-entrancy, we can probably consolidate some of these fields into a single state object.

jsalaber: If it is deemed necessary it could also be an option to make a separate class that takes care of the reconciliation of the context

btthomas: > it could also be an option to make a separate class

Here's an exploration of that class. Let me know what y'all think: https://github.com/btthomas/kotlin-sdk/pull/1/changes

nicklasl:
Non-blocking and unrelated to this PR’s changes: 
While reviewing the lifecycle coordination, I noticed that provider init can emit `READY` while `onContextSet` is still running, temporarily replacing `RECONCILING`. This behaviour predates this PR, so I don’t think it should expand the scope or block merging. I'll open a separate issue to track it separately.

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/OpenFeatureAPIInstance.kt
nicklasl: After looking through the concurrency cases a bit, I think supporting overlapping reconciliation is worthwhile. `setEvaluationContext()` cancels the previous `setEvaluationContextJob` without awaiting its completion, while concurrent callers can also invoke `setEvaluationContextAndWait()`. A providers `onContextSet()` implementation might find it hard to respond to cancellation promptly.
Holding a mutex across `onContextSet()` could therefore allow stale provider work to block the latest call to `setEvaluationContext()` or `setEvaluationContextAndWait()`. I'm concerned it may also risk deadlocking if an `onContextSet()` implementation re-enters the API by calling `setEvaluationContextAndWait()`.
 
My take is that the reconciliation bookkeeping seems justified, although I agree it might not belong naturally in `OpenFeatureAPIInstance`. The extracted `ContextReconciler` looks like a good follow-up refactor. I'm comfortable keeping that separate so this PR remains focused on behaviour and its tests.

nicklasl (approved):

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/OpenFeatureAPIInstance.kt
```diff
@@ -224,16 +240,99 @@ open class OpenFeatureAPIInstance internal constructor() {
...
+
+    private suspend fun completeContextReconciliation(
+        reconciliationProvider: FeatureProvider,
+        reconciliationProviderGeneration: Long,
+        terminalStatus: OpenFeatureStatus?
+    ) {
+        contextReconciliationMutex.withLock {
+            if (contextReconciliationGeneration != reconciliationProviderGeneration) return
+
+            if (terminalStatus != null) {
```
mtonko-flx, line 309: @btthomas Hey, Blake! I'm new contributor to the project, so please take my comments with a grain of salt :)

Lock here serializes the write but `onContextSet` still [runs outside it](https://github.com/open-feature/kotlin-sdk/pull/251/changes#diff-e14506f7c94b4d277a6f05c246ed659d7204ff72630fc43ef912ea8474c028bfR274), so two overlapping `setEvaluationContextAndWait` calls can finish out of order - whichever one completes last overwrites terminal status, even if that’s the older context. 

"Last finish wins" behavior already existed on `main` for unequal contexts, but equal contexts now overlap too. We stay `Reconciling` until everything finishes, then publish whichever `onContextSet` happened to complete last. Two refreshes of the same context can still disagree (one succeeds, one times out). I'd assume that the status we keep should just come from the most recently registered context, not whichever call finishes last.

Maybe not completely a subject of this PR specifically but wanted to highlight. WDYT?

toddbaert (approved):

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/OpenFeatureAPIInstance.kt
nicklasl: I think this sounds very much related to what I'm highlighting [here](https://github.com/open-feature/kotlin-sdk/pull/251#issuecomment-5326216842) so I decided to make #254 cover both of these cases so we can resolve this comment and merge the PR.

PR #253, opened 2026-08-12
fix: snapshot provider and context for track and evaluation

### LINKED ISSUE #252: track() and flag evaluation can use mismatched provider/context pairs
## Problem

`OpenFeatureClient.track()` and `evaluateFlag()` read the current provider and evaluation context via two separate calls:

```kotlin
openFeatureAPI.getProvider()
    .track(trackingEventName, openFeatureAPI.getEvaluationContext(), details)
```

`getProvider()` and `getEvaluationContext()` are plain field reads with no shared synchronization. Another thread can swap the provider or update context between those reads, producing calls like `ProviderA.track(..., contextForProviderB, ...)`.

The static-context client API is already correct (`track(name, details?)` only — no per-call context overloads). This is a provider/context **consistency** bug, similar to what was fixed in [open-feature/swift-sdk#124](https://github.com/open-feature/swift-sdk/pull/124).

## Impact

- Incorrect tracking attribution when provider/context change concurrently
- Same two-read pattern exists in `evaluateFlag()`, so flag evaluation can suffer the same mismatch

## Proposed fix

Introduce an internal atomic snapshot (e.g. `getEvaluationState()`) that reads `(provider, context)` under one lock, and use it from `track()` and `evaluateFlag()`. Guard all provider/context writes with the same lock.

## Related

- [open-feature/swift-sdk#123](https://github.com/open-feature/swift-sdk/issues/123) — static-context tracking API (Kotlin does not have this part of the issue)
- [open-feature/swift-sdk#124](https://github.com/open-feature/swift-sdk/pull/124) — provider/context snapshot fix in Swift SDK

fabriziodemaria:
## Summary

- Add `getEvaluationState()` to read `(provider, context)` under a single lock
- Use the snapshot in `OpenFeatureClient.track()` and `evaluateFlag()` instead of separate `getProvider()` + `getEvaluationContext()` calls
- Guard all provider/context writes with the same lock
- Add regression tests, including a concurrent swap/track test

Fixes #252

## Context

The Kotlin SDK already exposes the correct static-context tracking API (`track(name, details?)` only). The bug is provider/context **consistency**: two unsynchronized field reads could interleave with concurrent provider swaps or context updates.

Same class of issue fixed in [open-feature/swift-sdk#124](https://github.com/open-feature/swift-sdk/pull/124).

## Test plan

- [x] `:kotlin-sdk:jvmTest` (includes new `EvaluationStateSnapshotTests`)
- [ ] CI `check` (full multiplatform + Android — requires CI environment)

Made with [Cursor](https://cursor.com)


kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/OpenFeatureAPIInstance.kt
```diff
@@ -142,7 +147,8 @@ open class OpenFeatureAPIInstance internal constructor() {
             // Initialize the new provider
             tryWithStatusEmitErrorHandling {
                 listenToProviderEvents(provider, dispatcher)
-                getProvider().initialize(context)
+                val initialEvaluationContext = synchronized(stateLock) { context }
+                provider.initialize(initialEvaluationContext)
```
vahidlazio: previously we used `getProvider` but not use provider, is that intentional?

fabriziodemaria: Good catch — not intentional. Switched to `getEvaluationState()` so `initialize` reads provider and context from the same snapshot, consistent with `track`/`evaluateFlag`. Pushed in 581267e.

vahidlazio (approved):

PR #258, opened 2026-08-28
feat: complete the provider event vocabulary

### LINKED ISSUE #365: Provider lifecycle race condition in multi-threaded SDKs
I've been digging provider lifecycle in Go and Swift SDKs, and I believe there's a couple of things in OF spec that make it hard to implement it correctly in the presence of true multi-threading.

The main one is that there's no clear ownership of the provider state and provider events — it's split between the provider and the API, and both provider and the API are allowed to emit events. This is an issue because neither the provider nor the API are equipped to ensure the consistent ordering of events

The API is required to emit "provider ready" event and update the state after initialization is complete. However, once the provider returns from the init function, there's a brief window before the API updates the state and emits the event. This is not an issue in single-threaded languages like js or python. However, with multi-threading, the provider is allowed to do work in the background and emit an error event within that window — and now the order of provider ready and error events is undefined.

Current state across multi-threaded SDKs:
- Go: race condition
- Swift: 🐛 implements eventing but does not update state on provider events
- Java: race condition
- Kotlin: race condition
- Dotnet: race condition
- Rust: eventing not implemented

This would be trivial to fix if only one of API or provider were allowed to emit events (or update the state), so they could use locks/queues to order the events/updates properly.

If both are allowed to emit, the only way to prevent a race is for the API to guard eventing with a lock and to hold this lock throughout the provider initialization. This way, no provider events would come through until API processes its own ready event first.

The main issue with this approach is that it locks up the API/provider and serializes all calls to init/shutdown/on context change. This more or less completely destroys concurrency and prevents the provider from reacting to context change / shutdown if initialization is taking a while. This also easily leads to deadlocks if provider tries to emit events from/during lifecycle handlers.

Both race condition and total locking sound bad, so I would really suggest moving state management responsibility back to providers — they are the only party who knows their own state at all times, and they are equipped to resolve any races.

I see that the spec is trying to lift the burden from provider authors but imo SDK trying to manage providers' state is not helping. Anecdotally, I can add that as a provider author, the eventing and state management has been the most confusing part of the OF spec/SDKs, and I've spent quite a bit of time trying to make SDK set the state I want.

--- 

PoCs by @toddbaert here:

Go: https://github.com/open-feature/go-sdk/pull/487
Java: https://github.com/open-feature/java-sdk/pull/1892
JS: https://github.com/open-feature/js-sdk/pull/1362

mtonko-flx:
## Intent

The provider event vocabulary is missing two of its members. `PROVIDER_RECONCILING` and `PROVIDER_CONTEXT_CHANGED` are part of the event set the specification requires, but no corresponding types exist, so the SDK has no way to represent context reconciliation as events. Event details also carry no provider name, so an event cannot be attributed to the provider that emitted it. This PR adds both, plus the event/status mapping as a single function, and nothing consumes them yet.

## Motivation

This is the groundwork for adopting [spec #385](https://github.com/open-feature/spec/pull/385) (*provider state ownership via events*, with clarifications in [#408](https://github.com/open-feature/spec/pull/408)), which resolves [spec #365](https://github.com/open-feature/spec/issues/365) — the provider lifecycle race in multi-threaded SDKs. Under that change the SDK derives provider status entirely from provider-emitted events, which is not expressible until the event set is complete and there is one authoritative event → status mapping.

Kept deliberately additive so the vocabulary can be reviewed on its own, separately from the behavioural change that starts using it.

## Spec Requirements

| --- | --- |
| [5.1.1](https://openfeature.dev/specification/sections/events/#requirement-511) — the event set **MUST** include `PROVIDER_READY`, `PROVIDER_ERROR`, `PROVIDER_CONFIGURATION_CHANGED`, `PROVIDER_STALE`, `PROVIDER_RECONCILING`, `PROVIDER_CONTEXT_CHANGED` | Satisfied — the two missing members are added |
| [5.2.3](https://openfeature.dev/specification/sections/events/#requirement-523) — "The `event details` **MUST** contain the `provider name` associated with the event" | Introduced (the field); populated by the SDK in the follow-up |
| [5.3.5](https://openfeature.dev/specification/sections/events/#requirement-535) — event → status association table | Introduced as code; applied in the follow-up |

### Implementation
- Adds `OpenFeatureProviderEvents.ProviderReconciling` and `ProviderContextChanged`.
- Adds `EventDetails.providerName`, defaulted so existing construction sites are unaffected.
- Adds `toOpenFeatureStatus()`, expressing the requirement 5.3.5 table once, reusing the existing `toOpenFeatureStatusError()` for the `ERROR`/`FATAL` split.
- Replaces `MultiProvider`'s inline event → status `when` with that shared function. Extending the sealed class made the existing `when` non-exhaustive; using the shared mapping rather than adding an `else` branch means the two new events participate in aggregation instead of being silently dropped.

## Testing

- Every row of the requirement 5.3.5 table, including that `PROVIDER_CONFIGURATION_CHANGED` implies no status transition and that a `PROVIDER_FATAL` error code maps to `FATAL` rather than `ERROR`.
- `EventDetails` carries a provider name.
- Existing suites unchanged: 234 tests, `ktlintCheck` and `apiCheck` green, `apiDump` regenerated.

## Breaking Changes

None — additive. `EventDetails` gains a defaulted parameter, so the public API grows without changing existing behaviour.

### DIFF
--- a/kotlin-sdk/api/android/kotlin-sdk.api
+++ b/kotlin-sdk/api/android/kotlin-sdk.api
@@ -694,19 +694,21 @@ public abstract class dev/openfeature/kotlin/sdk/events/OpenFeatureProviderEvent
 
 public final class dev/openfeature/kotlin/sdk/events/OpenFeatureProviderEvents$EventDetails {
 	public fun <init> ()V
-	public fun <init> (Ljava/util/Set;Ljava/lang/String;Ldev/openfeature/kotlin/sdk/exceptions/ErrorCode;Ljava/util/Map;)V
-	public synthetic fun <init> (Ljava/util/Set;Ljava/lang/String;Ldev/openfeature/kotlin/sdk/exceptions/ErrorCode;Ljava/util/Map;ILkotlin/jvm/internal/DefaultConstructorMarker;)V
+	public fun <init> (Ljava/util/Set;Ljava/lang/String;Ldev/openfeature/kotlin/sdk/exceptions/ErrorCode;Ljava/util/Map;Ljava/lang/String;)V
+	public synthetic fun <init> (Ljava/util/Set;Ljava/lang/String;Ldev/openfeature/kotlin/sdk/exceptions/ErrorCode;Ljava/util/Map;Ljava/lang/String;ILkotlin/jvm/internal/DefaultConstructorMarker;)V
 	public final fun component1 ()Ljava/util/Set;
 	public final fun component2 ()Ljava/lang/String;
 	public final fun component3 ()Ldev/openfeature/kotlin/sdk/exceptions/ErrorCode;
 	public final fun component4 ()Ljava/util/Map;
-	public final fun copy (Ljava/util/Set;Ljava/lang/String;Ldev/openfeature/kotlin/sdk/exceptions/ErrorCode;Ljava/util/Map;)Ldev/openfeature/kotlin/sdk/events/OpenFeatureProviderEvents$EventDetails;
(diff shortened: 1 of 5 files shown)


mtonko-flx:

PR #259, opened 2026-08-28
fix!: derive provider status from a single ordered event relay

mtonko-flx:
> **Stacked on #258.** Base is `main` because the branch lives in a fork and GitHub requires a base branch in the upstream repository, so the diff shown here includes #258's commit. Review only the second commit, `fix!: derive provider status from a single ordered event relay`, or merge #258 first and this diff will reduce to it.

## Intent

Provider status and provider events are two independent subscriptions to the same provider stream. `handleProviderEvents` collects it to update the status, and `observe<T>()` hands application code its own subscription — so nothing orders the two, and a handler can observe an event before the status reflects it. SDK-inferred transitions also never appear in the event stream at all. This PR makes the SDK collect the provider once, write the status, and only then republish, so the status is always current by the time subscribers see the event that caused it.

## Motivation

Requirement 5.3.5 requires the status be updated *before* the handlers for that event run; with two independent subscriptions that cannot be guaranteed. Requirement 5.3.3 requires a handler attached when the provider is already in a state to run immediately, which nothing currently provides. Both are prerequisites for [spec #385](https://github.com/open-feature/spec/pull/385), where the status is derived entirely from provider events.

It also closes the gap raised on [spec #365](https://github.com/open-feature/spec/issues/365) that Kotlin's `observe` is a straight passthrough to the provider, so lifecycle events around `setProvider`/`setEvaluationContext` never appear in it.

## Spec Requirements

| --- | --- |
| [5.3.5](https://openfeature.dev/specification/sections/events/#requirement-535) — the SDK **MUST** update the status "**before** invoking any event handlers for that event" | **Fixes an existing violation** |
| [5.3.3](https://openfeature.dev/specification/sections/events/#requirement-533) — "Handlers attached after the provider is already in the associated state, **MUST** run immediately" | **Fixes an existing violation** |
| [5.2.3](https://openfeature.dev/specification/sections/events/#requirement-523) — event details **MUST** contain the provider name | **Fixes an existing violation** — stamped by the relay |
| [1.7.1](https://openfeature.dev/specification/sections/flag-evaluation/#requirement-171), [1.7.2.1](https://openfeature.dev/specification/sections/flag-evaluation/#conditional-requirement-1721) | Satisfied — status accessor incl. `RECONCILING` |
| [1.7.3](https://openfeature.dev/specification/sections/flag-evaluation/#requirement-173), [1.7.4](https://openfeature.dev/specification/sections/flag-evaluation/#requirement-174), [1.7.5](https://openfeature.dev/specification/sections/flag-evaluation/#requirement-175) | Satisfied — derived from provider events via the 5.3.5 table |
| [1.7.6](https://openfeature.dev/specification/sections/flag-evaluation/#requirement-176) — `NOT_READY` once `shutdown` terminates | Satisfied — inferred by the SDK, status only, no event (5.1.1 has no `PROVIDER_NOT_READY`) |
| [5.2.6](https://openfeature.dev/specification/sections/events/#requirement-526) — handlers **MUST** persist across provider changes | **Improved** — an SDK-owned stream outlives provider swaps |
| [1.4.9](https://openfeature.dev/specification/sections/flag-evaluation/#requirement-149), [1.4.10](https://openfeature.dev/specification/sections/flag-evaluation/#requirement-1410) | Preserved — see *NOT_READY/FATAL short-circuit* below |

### Implementation
- One internal collector on `provider.observe()`: stamps the provider name, writes the status from the event, then republishes onto an SDK-owned `MutableSharedFlow` that `observe<T>()` reads.
- Republication uses `tryEmit` against a bounded buffer, so a slow application subscriber can never stall the SDK's own status derivation — requirement 1.1.2.4 notes application handlers need not complete. Overflow drops the oldest and logs.
- New subscribers receive an event synthesised from the *current status*, not a replay of the last raw event. Replaying the last event would resurface a stateless `PROVIDER_CONFIGURATION_CHANGED`, or an event belonging to a provider that has since been replaced.
- `statusFlow` becomes a `StateFlow`; `getStatus()` reads `.value` instead of a replay cache.

### NOT_READY/FATAL short-circuit
Requirements 1.7.6/1.7.7 as they stood — the client short-circuiting evaluation while `NOT_READY`/`FATAL` — were **removed** by #385 and retagged `@spec-2.2.7`, on the grounds that reading the status and then evaluating is itself a time-of-check/time-of-use gap ([spec #369](https://github.com/open-feature/spec/issues/369)). They were removed rather than inverted, so short-circuiting remains *permitted*. This SDK keeps it: dropping it would change the error codes callers already depend on, and `EvaluationState` snapshots the provider and context together, which bounds the staleness. Only the KDoc changes, to record that this is now deliberate policy rather than a requirement.

## Testing

- A handler reads the matching status at the moment it runs (5.3.5).
- A subscriber attaching while the provider is already `READY` — and separately while `STALE` — receives that state's event immediately; a stateless `CONFIGURATION_CHANGED` is not resurfaced, nor is an event from a replaced provider (5.3.3).
- Every relayed event carries the emitting provider's name, synthesised ones included (5.2.3).
- `shutdown` reverts to `NOT_READY` without putting an event into the stream (1.7.6).
- 241 tests, `ktlintCheck` and `apiCheck` green, `apiDump` regenerated.

Two pre-existing test weaknesses surfaced and were corrected rather than worked around: several event tests launched a collector and then registered a provider before it had subscribed, passing only because the *provider's own* flow replayed; and status-sequence assertions relied on the relay running in real time. Both now subscribe first and drive the relay on the test dispatcher.

## Breaking Changes

`statusFlow` is now `StateFlow<OpenFeatureStatus>` rather than `Flow<OpenFeatureStatus>`, and conflates consecutive equal values — status is a snapshot, and the transition sequence belongs to `observe<T>()`. Source-compatible for collection; binary-incompatible, and code asserting on every intermediate status should move to the event stream.

### DIFF
--- a/kotlin-sdk/api/android/kotlin-sdk.api
+++ b/kotlin-sdk/api/android/kotlin-sdk.api
@@ -286,10 +286,11 @@ public class dev/openfeature/kotlin/sdk/OpenFeatureAPIInstance {
 	public final fun getEvaluationContext ()Ldev/openfeature/kotlin/sdk/EvaluationContext;
 	public final fun getHooks ()Ljava/util/List;
 	public final fun getProvider ()Ldev/openfeature/kotlin/sdk/FeatureProvider;
+	public final fun getProviderEvents ()Lkotlinx/coroutines/flow/Flow;
 	public final fun getProviderMetadata ()Ldev/openfeature/kotlin/sdk/ProviderMetadata;
 	public final fun getProvidersFlow ()Lkotlinx/coroutines/flow/MutableStateFlow;
 	public final fun getStatus ()Ldev/openfeature/kotlin/sdk/OpenFeatureStatus;
-	public final fun getStatusFlow ()Lkotlinx/coroutines/flow/Flow;
+	public final fun getStatusFlow ()Lkotlinx/coroutines/flow/StateFlow;
 	public final fun setEvaluationContext (Ldev/openfeature/kotlin/sdk/EvaluationContext;Lkotlinx/coroutines/CoroutineDispatcher;)V
 	public static synthetic fun setEvaluationContext$default (Ldev/openfeature/kotlin/sdk/OpenFeatureAPIInstance;Ldev/openfeature/kotlin/sdk/EvaluationContext;Lkotlinx/coroutines/CoroutineDispatcher;ILjava/lang/Object;)V
 	public final fun setEvaluationContextAndWait (Ldev/openfeature/kotlin/sdk/EvaluationContext;Lkotlin/coroutines/Continuation;)Ljava/lang/Object;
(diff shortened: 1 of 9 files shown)


mtonko-flx:

PR #262, opened 2026-08-28
fix!: move provider status tracking back to provider

mtonko-flx:
## Intent

The SDK derived `OpenFeatureStatus` from what a provider *did* — returning, throwing, or staying silent — while the provider emitted its own events. Two sources of truth for one piece of state left a window after `initialize` completed but before the SDK updated state, in which a provider's events had no defined order against the SDK's, and made the client's pre-evaluation status check unreliable by construction.

This ports [open-feature/swift-sdk#114](https://github.com/open-feature/swift-sdk/pull/114) (`fix!: move provider status tracking back to provider`) to Kotlin: the provider owns `status` and reports every transition as an event; the SDK reads status, republishes events, and infers nothing. It closes the same three upstream issues:

- [spec#365](https://github.com/open-feature/spec/issues/365) — provider lifecycle race from split status ownership.
- [spec#369](https://github.com/open-feature/spec/issues/369) — TOCTOU on the client's status check. This was **former** requirements 1.7.6 and 1.7.7, removed from the specification in [spec#385](https://github.com/open-feature/spec/pull/385) (v0.9.0). Not to be confused with today's 1.7.6, which is the unrelated "not-ready once `shutdown` terminates".
- [swift-sdk#108](https://github.com/open-feature/swift-sdk/issues/108) — lifecycle over-serialization.

### Implementation

- **New `ProviderStatusTracker`** — public, provider-side. Derives status from the events sent through it, republishes them, and replays the current status to a new subscriber. `reconciling { }` reports the transitions around a reconciliation, collapsing overlapping invocations into one reported outcome per requirements 5.3.4.2/5.3.4.3.
- **`FeatureProvider`** gains an abstract `status`; `observe()` loses its default. `shutdown()` is retained (the Swift protocol has none) with a contract to return to not-ready if the provider can be registered again.
- **`OpenFeatureAPIInstance`** keeps no status. `getStatus()` and `statusFlow` both project `FeatureProvider.status`. The one inferred transition is shutdown, where requirement 1.7.6 makes it the SDK's own conclusion: `clearProvider` installs a provider that was never initialized.
- **Lifecycle dispatch** runs on the registration's own serial dispatcher, so calls are entered in order without the SDK waiting for one to finish before entering the next. A throw is no longer a status signal — it is logged, and the status stays where the provider put it. Cancellation propagates, and an awaited call waits for the provider to finish unwinding so the outcome it owes a reconciliation is reported first.
- **`OpenFeatureClient`** no longer checks status before evaluating. Evaluations always reach the provider and its error surfaces through the normal hook lifecycle. `EvaluationState` now carries hooks, so an evaluation takes one atomic snapshot instead of two reads.
- **`MultiProvider`** owns a tracker and reads children's statuses directly, so its bookkeeping map, status flow and event flow are gone. `Strategy` gains `status(providers)` with a default implementation.
- **`OpenFeatureStatus.Error`/`Fatal`** compare by the failure they describe. Without that, the same failure reported live and replayed to a late subscriber is two distinct statuses that `distinctUntilChanged` cannot collapse.
- **Two new events**, `ProviderReconciling` and `ProviderContextChanged`, plus the specification's event/status association table as shared helpers.

### Usage Examples

```kotlin
class MyProvider : FeatureProvider {
    private val statusTracker = ProviderStatusTracker()

    override val status: OpenFeatureStatus get() = statusTracker.status
    override fun observe(): Flow<OpenFeatureProviderEvents> = statusTracker.observe()

        connect(initialContext)
        statusTracker.send(OpenFeatureProviderEvents.ProviderReady())
    }

    override suspend fun onContextSet(
        oldContext: EvaluationContext?,
        newContext: EvaluationContext
    ) = statusTracker.reconciling { refresh(newContext) }

    override fun shutdown() = statusTracker.reset()

    fun onConnectionLost() = statusTracker.send(OpenFeatureProviderEvents.ProviderStale())
}
```

Application code is unchanged:

```kotlin
OpenFeatureAPI.setProviderAndWait(MyProvider())
OpenFeatureAPI.getStatus()                           // now reads the provider
OpenFeatureAPI.statusFlow.collect { … }              // still available, now derived
OpenFeatureAPI.observe<OpenFeatureProviderEvents>()  // still available
```

## Testing

`./gradlew clean check` green on macOS across jvm, android debug + release, iosSimulatorArm64, js/node and js/browser.

- `ProviderStatusTrackerTests` — 22 tests over the event/status table, the replay contract (nothing replayed while not-ready, replayed once otherwise, per-subscriber, delivered before live events with no gap or duplicate, stateless events delivered but never replayed), status current by the time a subscriber sees the event, and the coalescing.
- `ProviderStatusTrackerConcurrencyTest`, JVM-only since JS and native are single-threaded — four tests racing subscribe against send over 500 iterations each: strict monotonicity (which forbids both a replay that repeats a live event and any reordering), the replay handing off to the live stream without loss, racing reconciliations never stranding the tracker, and status consistency under concurrent senders.
- Behaviour changes visible in existing tests, each now asserting the new truth: a provider that reconciles without reporting contributes no transition; a `ProviderConfigurationChanged` no longer clears an error; a double binding fails loudly instead of reporting an error status.
- `ProviderLifecycleTests` — the spec#365 contract (a `PROVIDER_STALE` reported *during* `initialize` is not overwritten when it returns), a provider that throws without reporting staying `NotReady`, one that reports and then throws keeping what it reported, a provider superseded before its registration ever ran still being shut down and released, and a same-instance rebind not shutting the provider down.
- `MultiProviderTests` — a cancelled context set not stranding the aggregate at `Reconciling`, overlapping context sets reporting `Reconciling` once and only the last outcome, an error aggregate carrying the triggering child's `flagsChanged` and `eventMetadata`, a child failing to initialize not cancelling its siblings, and an aggregate returning to `NotReady`.

## Breaking Changes

- `FeatureProvider` gains an abstract `status`, and `observe()` is no longer defaulted, so **every provider must be updated**. The migration is four lines: hold a `ProviderStatusTracker`, delegate `status` and `observe()` to it, report an outcome from `initialize`, and reset it from `shutdown`. There is no compatibility path — a provider that reports nothing stays `NotReady`.
- `initialize` and `onContextSet` failures are reported by emitting `ProviderError`, not by throwing; the SDK no longer converts a throw into a status.
- The client no longer short-circuits on `NotReady` or `Fatal`; callers see whatever the provider returns or throws.
- `OpenFeatureAPIInstance.providersFlow` is removed, and `observe()` is now an unparameterised member with a reified extension alongside it.
- `MultiProvider.statusFlow` is removed in favour of `MultiProvider.status`. `Strategy` gains `status(providers)`, which is defaulted, so existing strategies need no change.
- `setProviderAndWait`'s `dispatcher` now defaults to the caller's, and **`setEvaluationContext`'s `dispatcher` parameter is removed** — a registration's own dispatcher runs its reconciliations so they are ordered against its `initialize`.
- `OpenFeatureStatus.Error`/`Fatal` now define `equals`/`hashCode`, comparing by the failure they describe.
- An aggregate `MultiProvider` status of `NOT_READY` reaches `getStatus()` but not `observe()`/`statusFlow`: the specification has no `PROVIDER_NOT_READY` event to carry it. The Swift implementation has the same gap.
- Re-registering a provider that is already registered keeps the dispatcher it was first registered with, because its lifecycle calls have to stay ordered against each other on one dispatcher.
- `setProvider` no longer waits for the outgoing provider's `shutdown`; `setProviderAndWait` and `clearProvider` do.
- A reconciliation begun while the provider is `NOT_READY` reports no events at all, where it previously concluded `READY` on success.

## Known gaps, deliberately left

- `MultiProvider.updateStatus` reuses `statusTracker.reset()` to mean "the aggregate became not-ready", which also bumps the reconciliation generation and clears its restore point. Correct today, but it couples two concerns; a private "establish status without an event" path would say what is meant.
- For a child event carrying no status, the re-aggregation runs without a trigger, so a configuration change that also moves the aggregate loses `flagsChanged`/`eventMetadata`.
- `MultiProvider.watchScope` is an unsynchronized `var` touched from `initialize` and `shutdown`, so a swap racing a shutdown can leave a watch scope running.
- A child that throws from `initialize`/`onContextSet` without emitting a `ProviderError` leaves no trace: `MultiProvider` has no logger, where the top-level equivalent logs exactly that case.

## Recommended review order

1. `events/OpenFeatureProviderEvents.kt` — the two new events and the event/status table
2. `ProviderStatusTracker.kt` — the new component; read `observe()` closely, the sequence fence is what makes replay atomic
3. `FeatureProvider.kt` — the contract change
4. `OpenFeatureAPIInstance.kt` — what was deleted, and how `statusFlow` is derived
5. `OpenFeatureClient.kt` — the TOCTOU removal
6. `multiprovider/MultiProvider.kt` — aggregate status and the precedence divergence
7. `NoOpProvider.kt` — the smallest complete provider
8. `ProviderStatusTrackerTests.kt`, `jvmTest/ProviderStatusTrackerConcurrencyTest.kt`
9. Updated test helpers, then the updated existing tests
10. `README.md`, `docs/multiprovider/README.md`

### DIFF
--- a/kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/FeatureProvider.kt
+++ b/kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/FeatureProvider.kt
@@ -3,30 +3,99 @@ package dev.openfeature.kotlin.sdk
 import dev.openfeature.kotlin.sdk.events.OpenFeatureProviderEvents
 import dev.openfeature.kotlin.sdk.exceptions.OpenFeatureError
 import kotlinx.coroutines.flow.Flow
-import kotlinx.coroutines.flow.emptyFlow
 import kotlin.coroutines.cancellation.CancellationException
 
+/**
+ * The interface implemented by upstream flag providers to resolve flags for their service.
+ *
+ * A provider is responsible for its own [status] and for emitting the events that explain it. The
+ * easiest way to satisfy both is to delegate to [ProviderStatusTracker]:
+ *
+ * ```kotlin
+ * class MyProvider : FeatureProvider {
+ *     private val statusTracker = ProviderStatusTracker()
+ *
+ *     override val status: OpenFeatureStatus get() = statusTracker.status
+ *     override fun observe(): Flow<OpenFeatureProviderEvents> = statusTracker.observe()
+ *
+ *     override suspend fun initialize(initialContext: EvaluationContext?) {
+ *         connect(initialContext)
+ *         statusTracker.send(OpenFeatureProviderEvents.ProviderReady())
+ *     }
+ *
+ *     override suspend fun onContextSet(
+ *         oldContext: EvaluationContext?,
+ *         newContext: EvaluationContext
+ *     ) = statusTracker.reconciling { refresh(newContext) }
+ *
+ *     override fun shutdown() = statusTracker.reset()
+ *
+ *     // ... flag evaluation methods ...
+ * }
+ * ```
+ *
+ * [ProviderStatusTracker] keeps [status] in sync with the events sent through it, is safe to use
+ * from any thread, and replays the current status to a new subscriber.
+ */
 interface FeatureProvider {
     val hooks: List<Hook<*>>
     val metadata: ProviderMetadata
 
     /**
-     * Called by OpenFeatureAPI whenever the new Provider is registered
-     * This function should block until ready and throw exceptions if it fails to initialize
+     * The current lifecycle status of this provider.
+     *
+     * The provider alone is responsible for keeping this up to date. It must be
+     * [OpenFeatureStatus.NotReady] before [initialize] is called, and must reflect the most recently
+     * emitted event thereafter.
+     *
+     * This must be **thread-safe**: the SDK reads it from flag evaluation paths on any thread. The
+     * recommended way to satisfy that is to expose [ProviderStatusTracker.status] directly.
+     */
+    val status: OpenFeatureStatus
+
+    /**
+     * Called by OpenFeatureAPI when this provider is registered, to do whatever asynchronous setup it
+     * needs — fetching an initial flag configuration, opening a streaming connection.
+     *
+     * **Required status transition:** emit at least one event before returning, so that [status]
+     * moves away from [OpenFeatureStatus.NotReady]. [OpenFeatureProviderEvents.ProviderReady] on
+     * success and [OpenFeatureProviderEvents.ProviderError] on failure are the usual outcomes, but
+     * any other non-not-ready status is valid. Emitting before returning is what lets a caller of
+     * `setProviderAndWait` see the right status immediately.
+     *
+     * Throwing does **not** set a status: the SDK logs the failure and leaves [status] alone, so a
+     * provider that throws without emitting stays [OpenFeatureStatus.NotReady].
+     *
+     * Lifecycle calls are entered in the order they were made, but the SDK does not wait for one to
+     * finish before entering the next.
+     *
      * @param initialContext any initial context to be set before the provider is ready
      */
     @Throws(OpenFeatureError::class, CancellationException::class)
     suspend fun initialize(initialContext: EvaluationContext?)
 
     /**
-     * Called when the lifecycle of the OpenFeatureClient is over to release resources/threads
+     * Called when the lifecycle of the OpenFeatureClient is over to release resources/threads.
+     *
+     * A provider that can be registered again must return to [OpenFeatureStatus.NotReady] here —
+     * [ProviderStatusTracker.reset] does that — so a reused instance does not report the status it
+     * held before it was shut down.
      */
     fun shutdown()
 
     /**
      * Called by OpenFeatureAPI whenever the application sets the [EvaluationContext], including when
      * the new context is equal to or the same instance as the previous context.
-     * Implementations should suspend until the provider is ready again or throw an exception.
+     *
+     * Either return without emitting anything, where no reconciliation is needed, or emit
+     * [OpenFeatureProviderEvents.ProviderReconciling], do the work, and emit
+     * [OpenFeatureProviderEvents.ProviderContextChanged] or
+     * [OpenFeatureProviderEvents.ProviderError]. [ProviderStatusTracker.reconciling] does the latter,
+     * including collapsing overlapping invocations.
+     *
+     * Since the SDK does not wait for one lifecycle call to finish before entering the next, this can
+     * be entered while previous reconciliation work is still in flight; a provider reconciling
+     * asynchronously should handle that, for instance by cancelling the work it supersedes.
      *
      * @param oldContext The old EvaluationContext
      * @param newContext The new EvaluationContext
@@ -61,10 +130,10 @@ interface FeatureProvider {
     }
 
     /**
-     * Used by providers to expose internal events to the SDK or the application.
-     * This can be optionally implemented by the provider to expose a flow of internal events.
+     * The events this provider emits, for the SDK and the application.
+     *
+     * The SDK derives nothing from a provider's silence: every status transition must arrive here.
+     * Return [ProviderStatusTracker.observe] to get the required replay behaviour for free.
      */
-    fun observe(): Flow<OpenFeatureProviderEvents> {
-        return emptyFlow()
-    }
+    fun observe(): Flow<OpenFeatureProviderEvents>
 }
--- a/kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/FeatureProvider.kt
+++ b/kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/FeatureProvider.kt
@@ -8,63 +8,29 @@ import kotlin.coroutines.cancellation.CancellationException
 /**
  * The interface implemented by upstream flag providers to resolve flags for their service.
  *
- * A provider is responsible for its own [status] and for emitting the events that explain it. The
- * easiest way to satisfy both is to delegate to [ProviderStatusTracker]:
- *
- * ```kotlin
- * class MyProvider : FeatureProvider {
- *     private val statusTracker = ProviderStatusTracker()
- *
- *     override val status: OpenFeatureStatus get() = statusTracker.status
- *     override fun observe(): Flow<OpenFeatureProviderEvents> = statusTracker.observe()
- *
- *     override suspend fun initialize(initialContext: EvaluationContext?) {
- *         connect(initialContext)
- *         statusTracker.send(OpenFeatureProviderEvents.ProviderReady())
- *     }
- *
- *     override suspend fun onContextSet(
- *         oldContext: EvaluationContext?,
- *         newContext: EvaluationContext
- *     ) = statusTracker.reconciling { refresh(newContext) }
- *
- *     override fun shutdown() = statusTracker.reset()
- *
- *     // ... flag evaluation methods ...
- * }
- * ```
- *
- * [ProviderStatusTracker] keeps [status] in sync with the events sent through it, is safe to use
- * from any thread, and replays the current status to a new subscriber.
+ * A provider is responsible for its own [status] and for emitting the events that explain it, which
+ * [ProviderStatusTracker] does on its behalf. The README carries a worked example.
  */
 interface FeatureProvider {
     val hooks: List<Hook<*>>
     val metadata: ProviderMetadata
 
     /**
-     * The current lifecycle status of this provider.
-     *
-     * The provider alone is responsible for keeping this up to date. It must be
-     * [OpenFeatureStatus.NotReady] before [initialize] is called, and must reflect the most recently
-     * emitted event thereafter.
-     *
-     * This must be **thread-safe**: the SDK reads it from flag evaluation paths on any thread. The
-     * recommended way to satisfy that is to expose [ProviderStatusTracker.status] directly.
+     * The current lifecycle status of this provider, kept up to date by the provider alone. It must
+     * be [OpenFeatureStatus.NotReady] before [initialize] is called, must reflect the most recently
+     * emitted event thereafter, and must be thread-safe: the SDK reads it from flag evaluation paths
+     * on any thread.
      */
     val status: OpenFeatureStatus
 
     /**
-     * Called by OpenFeatureAPI when this provider is registered, to do whatever asynchronous setup it
-     * needs — fetching an initial flag configuration, opening a streaming connection.
+     * Called by OpenFeatureAPI when this provider is registered, to do whatever asynchronous setup
+     * it needs.
      *
-     * **Required status transition:** emit at least one event before returning, so that [status]
-     * moves away from [OpenFeatureStatus.NotReady]. [OpenFeatureProviderEvents.ProviderReady] on
-     * success and [OpenFeatureProviderEvents.ProviderError] on failure are the usual outcomes, but
-     * any other non-not-ready status is valid. Emitting before returning is what lets a caller of
-     * `setProviderAndWait` see the right status immediately.
-     *
-     * Throwing does **not** set a status: the SDK logs the failure and leaves [status] alone, so a
-     * provider that throws without emitting stays [OpenFeatureStatus.NotReady].
+     * Emit at least one event before returning, so that [status] moves away from
+     * [OpenFeatureStatus.NotReady] — usually [OpenFeatureProviderEvents.ProviderReady] or
+     * [OpenFeatureProviderEvents.ProviderError]. Throwing does not set a status: the SDK logs the
+     * failure and a provider that throws without emitting stays [OpenFeatureStatus.NotReady].
      *
      * Lifecycle calls are entered in the order they were made, but the SDK does not wait for one to
      * finish before entering the next.
@@ -77,9 +43,8 @@ interface FeatureProvider {
     /**
      * Called when the lifecycle of the OpenFeatureClient is over to release resources/threads.
      *
-     * A provider that can be registered again must return to [OpenFeatureStatus.NotReady] here —
-     * [ProviderStatusTracker.reset] does that — so a reused instance does not report the status it
-     * held before it was shut down.
+     * A provider that can be registered again must return to [OpenFeatureStatus.NotReady] here, so
+     * a reused instance does not report the status it held before it was shut down.
      */
     fun shutdown()
 
@@ -90,12 +55,12 @@ interface FeatureProvider {
      * Either return without emitting anything, where no reconciliation is needed, or emit
      * [OpenFeatureProviderEvents.ProviderReconciling], do the work, and emit
      * [OpenFeatureProviderEvents.ProviderContextChanged] or
-     * [OpenFeatureProviderEvents.ProviderError]. [ProviderStatusTracker.reconciling] does the latter,
-     * including collapsing overlapping invocations.
+     * [OpenFeatureProviderEvents.ProviderError]. [ProviderStatusTracker.reconciling] does the
+     * latter, including collapsing overlapping invocations.
      *
-     * Since the SDK does not wait for one lifecycle call to finish before entering the next, this can
-     * be entered while previous reconciliation work is still in flight; a provider reconciling
-     * asynchronously should handle that, for instance by cancelling the work it supersedes.
+     * This can be entered while previous reconciliation work is still in flight; a provider
+     * reconciling asynchronously should handle that, for instance by cancelling the work it
+     * supersedes.
      *
      * @param oldContext The old EvaluationContext
      * @param newContext The new EvaluationContext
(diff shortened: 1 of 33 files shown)


kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/OpenFeatureAPIInstance.kt
```diff
@@ -124,47 +132,59 @@ open class OpenFeatureAPIInstance internal constructor() {
...
                     this.provider = provider
-                    providerGeneration++
-                    providersFlow.value = provider
+                    this.providerRelay = relay
+                    this.relayScope = scope
+                    _providersFlow.value = provider
                     if (initialContext != null) context = initialContext
-                    current
+                    current to generation
                 }
```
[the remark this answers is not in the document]
mtonko-flx: Fixed

nicklasl:
Hey! I've started going through this with some AI-assisted analysis and wanted to share an early observation before getting into specifics.

One question I keep coming back to as I start working through this: is providerMutex intended to serialize only the provider swap, or the lifecycle transition as a whole? It's released before shutdownReplacedProvider(), relay start/initialization, and onContextSet(), so I’d like to understand what guarantees keep those operations ordered against another provider registration or context update. 

I'm still working through the PR, so consider this the opening of a conversation rather than a verdict. More to follow.

mtonko-flx:
[nicklasl](https://github.com/nicklasl) Hey! Thanks for feedback, please check again this PR. I took Swift implementation here https://github.com/open-feature/swift-sdk/pull/114 as a reference, though there are some differences present which I highlighted in PR description.

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/OpenFeatureAPIInstance.kt
```diff
@@ -1,357 +1,365 @@
...
  */
 @Suppress("TooManyFunctions")
 open class OpenFeatureAPIInstance internal constructor() {
-    private data class ContextReconciliation(
-        val oldContext: EvaluationContext?,
+    private val logger = LoggerFactory.getLogger(LOGGER_NAME)
+    private val stateLock = SynchronizedObject()
+
+    /** The provider installed when none has been registered, or once one has been cleared. */
+    private class NoProvider : NoOpProvider()
```
nicklasl, line 52: Could the fallback provider preserve the previous `PROVIDER_NOT_READY` evaluation semantics? With the client guard removed it inherits NoOpProvider’s successful default resolution, so evaluation details against this provider no error while `getStatus()` is `NotReady`.

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/OpenFeatureAPIInstance.kt
```diff
@@ -1,357 +1,365 @@
...
-                tryEmit(OpenFeatureStatus.NotReady)
+    private var registration = ProviderRegistration(NoProvider(), Dispatchers.Default)
+    private val providerRegistrations = MutableStateFlow(registration)
+
+    /** Never cancelled: a dropped retirement leaks the provider it was meant to release. */
+    private val retirementScope = CoroutineScope(
+        SupervisorJob() +
+            Dispatchers.Default +
+            CoroutineExceptionHandler { _, throwable ->
+                logger.warn({ "Retiring a replaced provider failed" }, throwable = throwable)
```
nicklasl, line 83: We should be careful about logging issues unless there is something that the SDK user can do differently to avoid the problem. 
Is this a case where the SDK user did something wrong or just a raise?

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/OpenFeatureAPIInstance.kt
```diff
@@ -380,74 +388,47 @@ open class OpenFeatureAPIInstance internal constructor() {
...
-        .flatMapLatest { it.observe() }.filterIsInstance()
+    fun observe(): Flow<OpenFeatureProviderEvents> =
+        providerRegistrations.flatMapLatest { it.provider.observe() }
 
     /**
-     * Aligning the state management to
-     * https://openfeature.dev/specification/sections/events#requirement-535
+     * Claims [provider] for this instance.
+     *
+     * @throws IllegalStateException if another instance already owns [provider]
```
nicklasl, line 427: Could we say "another OpenFeatureAPI instance" here instead of just instance. I was a bit confused :)

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/ProviderStatusTracker.kt
```diff
@@ -0,0 +1,224 @@
...
+ *
+ * A new subscriber receives one synthetic event reflecting the current status, so attaching does not
+ * race the first live event. Nothing is replayed while the provider is [OpenFeatureStatus.NotReady],
+ * which has no corresponding event type.
+ *
+ * A provider delegates [FeatureProvider.status] and [FeatureProvider.observe] to this. Do not
+ * collect [observe] on [kotlinx.coroutines.Dispatchers.Unconfined], and do not call [send] from
+ * inside a collector: delivery would then run inline under the lock that orders events.
+ */
+class ProviderStatusTracker {
```
nicklasl, line 48: Let's make it clear in the kDoc that this implementation will drop oldest events if they are slow to process and more than 64.

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/FeatureProvider.kt
```diff
@@ -3,30 +3,64 @@ package dev.openfeature.kotlin.sdk
...
 
     /**
-     * Called by OpenFeatureAPI whenever the new Provider is registered
-     * This function should block until ready and throw exceptions if it fails to initialize
+     * The current lifecycle status of this provider, kept up to date by the provider alone. It must
+     * be [OpenFeatureStatus.NotReady] before [initialize] is called, must reflect the most recently
+     * emitted event thereafter, and must be thread-safe: the SDK reads it from flag evaluation paths
+     * on any thread.
+     */
+    val status: OpenFeatureStatus
```
nicklasl, line 24: Can we get by without this accessor? I don't think it is required by the spec.

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/ProviderStatusTracker.kt
```diff
@@ -0,0 +1,224 @@
...
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.MutableSharedFlow
+import kotlinx.coroutines.flow.emitAll
+import kotlinx.coroutines.flow.filter
+import kotlinx.coroutines.flow.flow
+import kotlinx.coroutines.flow.map
+import kotlinx.coroutines.flow.onSubscription
+import kotlinx.coroutines.withContext
+
+private const val EVENT_BUFFER_CAPACITY = 64
```
nicklasl, line 21: If a provider emits more than 64 events before a subscriber gets a chance to process them DROP_OLDEST will silently discards earlier events, potentially including configuration changes. How should we preserve the delivery guarantee in [5.1.2](https://openfeature.dev/specification/sections/events/#requirement-512) when a subscriber falls behind?

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/OpenFeatureAPIInstance.kt
```diff
@@ -1,357 +1,365 @@
...
+     * its back, say — reports that through [getStatus] alone. A provider that reports nothing yields
+     * exactly one [OpenFeatureStatus.NotReady].
+     */
+    @OptIn(ExperimentalCoroutinesApi::class)
+    val statusFlow: Flow<OpenFeatureStatus> = providerRegistrations
+        .flatMapLatest { current ->
+            current.provider.observe()
+                .transform { event ->
+                    // The event's own status: a re-read loses the earlier of two transitions.
+                    val reported = event.toOpenFeatureStatus()
```
nicklasl, line 110: Could reading the live status here make us report transitions out of order? As I understand it, if the provider emits Reconciling → Ready → Stale before we process those events, this produces Reconciling → Stale → Ready → Stale, since the live status is already Stale. I guess we’ll need to derive each update from the event alone to preserve the sequence?

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/OpenFeatureAPIInstance.kt
mtonko-flx: Thanks, fixed. The fallback NoProvider now throws `ProviderNotReadyError` from evaluation methods, so an evaluation before a provider is set or after `clearProvider()` is back to `PROVIDER_NOT_READY`.

I left `track()` as the inherited no-op. It returns `Unit` with no error channel, so throwing could surprise callers rather than inform them and [tracking spec](https://openfeature.dev/specification/sections/tracking) doesn't tie tracking to status. Happy to change it if you'd rather it be explicit.

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/ProviderStatusTracker.kt
mtonko-flx: Fixed. The tracker's shared buffer is now unbounded (`extraBufferCapacity = Int.MAX_VALUE`), so a subscriber that falls behind delays its own delivery instead of losing events.

Worth mentioning about the trade with the cost of memory. A subscriber that never drains will grow the buffer until it OOMs, where before it silently lost the oldest events. That's why I initially made it with a cap but missed this spec requirement. But I think that is the right way round for [5.1.2](https://openfeature.dev/specification/sections/events/#requirement-512), because silent loss is undetectable and a stuck collector is a bug you can actually see

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/OpenFeatureAPIInstance.kt
mtonko-flx: Fixed

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/ProviderStatusTracker.kt
mtonko-flx: Fixed here: https://github.com/open-feature/kotlin-sdk/pull/262#discussion_r4062071288. It will not have a bound of 64 events.

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/FeatureProvider.kt
mtonko-flx: You are right that the spec does not require it and maybe even leans the other way:
> The SDK derives provider status from events emitted by the provider [...] the SDK updates its internal status accordingly", with shutdown as the exception the SDK infers itself.

However, [swift-sdk#114](https://github.com/open-feature/swift-sdk/pull/114) added the accessor and is merged and this PR is a port of it. Kotlin diverging quietly is worse, so decided to highlight this. Worth noting the PR's intent survives either way - the single source of truth is events. Scope if we take it back into the SDK: ~15 files declare override val status and it reshapes `getStatus()`, `statusFlow`, `MultiProvider` aggregation and every test helper providers.

Ready to do either way, I'll just need an advice with you about it here.

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/OpenFeatureAPIInstance.kt
mtonko-flx: Thanks, fixed

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/OpenFeatureAPIInstance.kt
mtonko-flx: I removed it and widened the catch in `retireProvider` to cover its whole body rather than just the `shutdown()` call, so nothing can escape to the scope by construction. The one log that remains is the "failed to shut down" warning, which is actionable: a provider whose `shutdown()` throws is a bug worth surfacing to whoever integrated it.

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/multiprovider/MultiProvider.kt
```diff
@@ -270,12 +336,24 @@ class MultiProvider(
...
+                        .awaitAll()
+                }
+                updateStatus()
+            }
+        } finally {
+            // Dropped only once the tracker has reported the outcome, which its own finally does
+            // before this one runs.
+            val settled = synchronized(statusLock) { --openReconciliations == 0 }
+            // Nothing else reports a recovery that landed while the reconciliation suppressed it.
+            if (settled) updateStatus()
```
nicklasl: When we `settle` for the last open reconciliation and the aggregate is `Ready`we should emit `ProviderContextChanged` in the `updateStatus()`.

This should be a pretty small thing, very compartmentalised to this class. I have an example commit: https://github.com/nicklasl/kotlin-sdk/commit/8c38b831e399622b281912dd05812060c66fedaf

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/OpenFeatureAPIInstance.kt
```diff
@@ -1,358 +1,405 @@
...
+            // Published under the lock that commits the swap, or two concurrent swaps can leave
+            // evaluations on one provider and statusFlow pinned to the other.
+            providerRegistrations.value = current
+            val initializationContext = context
+            val pendingRetirements = retirements.filter { it.first === newProvider }.map { it.second }
+            initialization = current.dispatchLifecycle("initialize", CoroutineStart.LAZY) {
+                // A retirement of this provider that already committed to shutting it down is still
+                // running: initializing over it would race its teardown.
+                pendingRetirements.forEach { it.join() }
+                newProvider.initialize(initializationContext)
```
nicklasl: Could we leave re-registering the same provider while it’s still shutting down unsupported in this PR? I’ve been looking at this with an AI, and my understanding is that this wait tries to support that reuse, but context changes can still reach the provider before reinitialization.

My take is that we should expect callers to wait for shutdown and providers to stop accepting new work during teardown, and avoid trying to coordinate overlapping reuse here.

The spec explicitly treats reinitialization as something _some providers support_, rather than a universal guarantee which is why I think we could leave it out, at least for now. [Requirement 2.5.2](https://openfeature.dev/specification/sections/providers/#requirement-252)

nicklasl: Here's a suggestion: https://github.com/nicklasl/kotlin-sdk/commit/7645f5f3a223dc8f0441c111aeac3977735ba42b

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/multiprovider/MultiProvider.kt
mtonko-flx: Thank you, fixed

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/OpenFeatureAPIInstance.kt
mtonko-flx: Thank you, used your commit as a reference and applied

nicklasl (approved):

kotlin-sdk/src/commonMain/kotlin/dev/openfeature/kotlin/sdk/FeatureProvider.kt
nicklasl: hmm... I can't argue with that keeping it doesn't has practical benefits (we use it for some stuff in the PR), but making every provider implement it is putting an API commitment on the `FeatureProvider` with the included state requirement of being able to respond to that API.

If we can get by without it, without taking on too much complexity in the SDK I would probably skip it even if the Swift SDK went with that approach.

Pinging @dd-oleksii who probably thought a lot about this.

